PR #52, opened 2026-08-04
release: derive acceptance counts from the shipping board set

jackspace:
The acceptance validators compute how many rows they require, then report a stale literal, so a
roster that fails is told to supply the wrong number.

`flasher_acceptance_contract.py:189` compares against `len(SHIPPING_BOARDS)`, which is five since
`heltec-v4-r8` landed, and then raises `exactly four well-formed targets`.
`flasher_tester_roster.py:231` compares against the `SHIPPING_BOARDS` by `SURFACES` product, ten,
and reports `exactly eight physical assignments`. `validate-flasher-tester-roster.py:44` prints the
same stale eight on success.

Each message now derives from the constant its own check already uses, so the two cannot drift
apart again. Rendered today:

```
manifest must contain exactly 5 well-formed targets
tester roster must contain exactly 10 physical assignments
tester roster covers 10 physical, 4 fallback, and 5 installer assignments
```

Four prose counts are corrected to match. One deliberate omission:
`docs/website/web-flasher/browser/README.md` still says four-board, because that is the pinned
0.2.6 fixture rather than the shipping matrix, so it is correct as written.

The duplicate-target test drops the count from its regex. It exercises duplicate rejection, not
arithmetic.

No behaviour change. Every check compared the same values before and after; only the text a human
reads is different.

Verified on this branch:

```
python -m unittest tools.tests.test_create_flasher_acceptance   ->  7 tests, OK
./tools/prns verify                                            ->  TOOLING_REGISTRY_OK
python validation/run.py verify                                ->  VALIDATION_REGISTRY_OK
```

Not run: the embedded and platform suites, since nothing here touches firmware or the site. I also
saw two pre-existing errors in `tools.tests.test_task_runner`
(`test_cargo_tools_is_a_thin_task_runner_alias` and `test_verify_output_explains_guarantees`) which
reproduce identically on unmodified trunk at the same commit on a Windows host, so they are not
from this change. Happy to open a separate issue with the output if that is useful.

### DIFF
--- a/docs/release-dependency-audit.md
+++ b/docs/release-dependency-audit.md
@@ -93,6 +93,6 @@ allowlist and cannot contain a runtime CDN or the legacy ESP Web Tools engine.
 - cargo-about 0.9.1
 - cargo-geiger 0.13.0 (advisory evidence)
 
-OS/hardware acceptance—including the complete web/CLI four-board matrix and every published CLI
+OS/hardware acceptance—including the complete web/CLI five-board matrix and every published CLI
 architecture—must be signed off against the exact signed candidate. The protected promotion job
 validates `acceptance.json`; those physical observations cannot be inferred from CI.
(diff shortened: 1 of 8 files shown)


KenAKAFrosty:
Love the diligence for single-source-of-truth. A nice, welcome change. Thanks!

PR #63, opened 2026-08-05
flash-manifest: read UF2 bootloader identity from the catalog instead of hardcoding it

jackspace:
Two commits. The first is a refactor with no behavior change for the T-Echo, and the second closes a gap the first one makes reachable.

I'm bringing up a second nRF52840 UF2 board and ran into both of these in the first hour, so I pulled them out on their own rather than bury them in a firmware change.

**The bootloader volume label.** `prns-flash-manifest/src/catalog.rs` validated every UF2 entry with `build.mount_label != "TECHOBOOT"`, which is a per-board catalog value checked against a hardcoded copy of itself. That's now a non-empty check. The strict label shape already lives where it needs to, at the untrusted bridge boundary in `docs/website/web-flasher/src/core.js`, and a second copy of that rule in Rust with nothing tying the two together is how they drift.

**The bootloader Board-ID prefix.** The flasher found the target drive by requiring `INFO_UF2.TXT`'s Board-ID to start with the literal `nrf52840-techo-v`. `Uf2Build` gains `board_id_prefix`, and a new `Uf2BoardIdPrefix` domain value owns the folding rule (lowercase, underscores to hyphens) and requires the catalog to store the already-folded form, so the trusted side of the comparison is never normalized at runtime. It's validated alongside `ChipFamily` and the reset strategies.

Detection has two named entry points now, and the difference is a safety property. `flash` narrows to the selected board's prefix. `doctor` with no board argument scans everything cataloged, because it has nothing to narrow to. The revision-shape rule is unchanged: the text after the prefix still has to be a non-empty alphanumeric-or-dot token, so a bare prefix, a generic UF2 drive, or a coincidental mount label still fails identity.

**The second commit.** While checking that narrowing claim I found it isn't airtight. Nothing compares prefixes across boards, and since detection strips a prefix and then requires a revision token after it, one board's prefix beginning with another's means a single mounted drive satisfies both boards' checks. Each prefix is individually well formed, so catalog validation is the only place to catch it. There's a new `OverlappingUf2BoardIdPrefixes` error and a test that clones the T-Echo entry with a longer overlapping prefix and expects a refusal. It can't bite today with one UF2 board cataloged, and it's exactly what a second one would spring.

What this deliberately doesn't touch: `PreparationProfile::TechoUf2`, since the wire spelling is baked into the minisign-signed 0.2.6 browser fixture; the `techo_mounts` field in schema-1 `doctor --json`, which is a public contract that should move with a schema bump; `HOPSPOT_TECHOBOOT`, which is an operator escape hatch; and `SHIPPING_BOARD_SLUGS`, because adding a board should stay a conscious edit.

Tested both ways, since I have a Windows box:

- Linux (WSL): `cargo test -p prns-flash-manifest -p hopspot-flash`, 126 tests across five binaries, 0 failed. `cargo clippy -p prns-flash-manifest -p hopspot-flash --all-targets -- -D warnings`, clean.
- Windows (x86_64-pc-windows-msvc): `cargo test -p prns-flash-manifest`, 38 passed, 0 failed.
- `python validation/run.py verify` and `./tools/prns verify` both green.

One thing I noticed from running it on both: `validate-flasher-candidate` runs 9 tests on Linux and 5 on Windows. The four that don't run are the symlink rejection ones, and they're Unix-gated. Nothing to do with this change, but it means those protections aren't exercised on Windows at all. Happy to open that separately if it's worth a look.

### DIFF
--- a/docs/website/src/pages/flash/model.rs
+++ b/docs/website/src/pages/flash/model.rs
@@ -176,7 +176,7 @@ pub(super) fn preparation_guide(
             ],
         },
         PreparationProfile::TechoUf2 => PreparationGuide {
-            lead: "T-Echo uses its UF2 bootloader; the website only verifies and downloads the UF2 file.",
+            lead: "This board uses its UF2 bootloader; the website only verifies and downloads the UF2 file.",
             steps: match target {
                 BoardFlashTarget::Uf2MassStorage { mount_label } => vec![
                     "Prepare the verified UF2 here before entering bootloader mode.".to_string(),
--- a/personal-hopspot/flasher/src/cli.rs
+++ b/personal-hopspot/flasher/src/cli.rs
@@ -95,7 +95,7 @@ pub(crate) enum CommandMode {
             conflicts_with_all = ["version", "offline", "local_build"]
         )]
         candidate: Option<PathBuf>,
-        /// Explicit mounted TECHOBOOT directory.
+        /// Explicit mounted UF2 bootloader directory.
         #[arg(long, value_name = "DIR", hide = true)]
         mount: Option<PathBuf>,
     },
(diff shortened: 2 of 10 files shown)


KenAKAFrosty:
Good catch once more!  Took it a bit further with the  SSOT and hardening. Great stuff, thanks

PR #65, opened 2026-08-05
release: bind the documented acceptance counts to the contract

jackspace:
Follow-up to #61, where you said pointing these at a single source of truth would be welcome.

`flasher_acceptance_contract.py` derives the qualification matrix. Five documents restate its sizes in prose, so a reader knows what they owe without reading Python. Those restatements do not update themselves: the four-board era's "eight rows" survived in two separate documents, and the T-Echo foundations change has since shown the same pattern again, with three more sentences going stale the moment the arithmetic changed.

This adds `validation/release/acceptance-doc-contracts.py`, which stores no count of its own. It derives every number from the contract and requires each governed sentence to state exactly that:

| quantity | derivation |
|---|---|
| physical rows | one per board/surface pair, times the S140 compatibility variants a UF2 target owes (transport from `release/flash/boards.json`, which must match `SHIPPING_BOARDS` exactly or the check refuses to vouch for the arithmetic) |
| roster assignments | `len(SHIPPING_BOARDS) * len(SURFACES)`, a smaller number than the rows since the T-Echo's two foundations share one assignment |
| browser fallbacks | `len(REQUIRED_FALLBACKS)` |
| fallback checkpoints | `len(FALLBACK_SCENARIOS)` |
| installer rows | `len(CLI_TARGETS)`, `len(OS_ARCHITECTURES)` |
| board matrix | `len(SHIPPING_BOARDS)` |

Adding a sixth board to `SHIPPING_BOARDS` turns every affected sentence red at once, across four documents, each naming the file and the exact sentence it expected:

```
acceptance doc contract: release/acceptance/README.md: expected the contract-derived sentence
"produces fourteen physical rows, four unsupported-browser rows, and five native installer rows".
```

A second pass then rejects any *unregistered* count in the same documents, so a new restatement cannot be added without being bound here too. It recognizes the subjects these documents already use; genuinely new phrasing for a new quantity still needs an entry, and the check says so.

Sentences are matched against whitespace-flattened text so a soft-wrapped sentence still matches, and errors still report the original line.

It runs inside the existing `release-contracts` suite, so there is no new suite or workflow. The five governed documents join that suite's declared inputs, so a docs-only edit reruns the check instead of hitting a cached pass.

Ran on Windows 11, `x86_64-pc-windows-msvc`:

- `bash validation/release/contracts.sh` - the new check passes inside the suite. The `tools/tests` phase reports 2 failures and 5 errors on this host; an unmodified worktree of trunk reports the same set, so they are pre-existing here and not from this change.
- Fail-closed verified three ways: a drifted governed count exits 1, an unregistered count exits 1, a `boards.json` that disagrees with the shipping set exits 1; the committed tree exits 0.

Rebased onto trunk `cf277157`, single commit.

### DIFF
--- a/tools/tests/test_acceptance_doc_contracts.py
+++ b/tools/tests/test_acceptance_doc_contracts.py
@@ -1,0 +1,118 @@
+from __future__ import annotations
+
+import importlib.util
+from pathlib import Path
+import tempfile
+import unittest
+from unittest.mock import patch
+
+
+ROOT = Path(__file__).resolve().parents[2]
+SCRIPT = ROOT / "validation" / "release" / "acceptance-doc-contracts.py"
+
+
+def load_checker():
+    spec = importlib.util.spec_from_file_location("acceptance_doc_contracts", SCRIPT)
+    if spec is None or spec.loader is None:
+        raise RuntimeError(f"could not import {SCRIPT}")
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+CHECKER = load_checker()
+
+
+class AcceptanceDocContractTests(unittest.TestCase):
+    def setUp(self) -> None:
+        self.temporary = tempfile.TemporaryDirectory()
+        self.root = Path(self.temporary.name)
+        self.path = self.root / "acceptance.md"
+        self.contracts = (
+            CHECKER.CountContract(
+                "acceptance.md",
+                "Acceptance requires ",
+                "physical",
+                " physical rows.",
+            ),
+        )
+        self.derived = {"physical": 12}
+
+    def tearDown(self) -> None:
+        self.temporary.cleanup()
+
+    def check(self) -> list[str]:
+        return CHECKER.check(self.root, self.derived, self.contracts)
+
+    def test_accepts_spelled_and_decimal_counts(self) -> None:
+        for count in ("twelve", "12"):
+            with self.subTest(count=count):
+                self.path.write_text(
+                    f"Acceptance requires {count} physical rows.\n",
+                    encoding="utf-8",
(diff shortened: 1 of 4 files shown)


jackspace:
Related, and a larger instance of the same thing: the acceptance contract's `REQUIRED_FALLBACKS` encodes a browser fact that expired when Firefox 151 shipped Web Serial. Details in #67.

jackspace:
Heads up that this one is half applied already. `git cherry trunk` against the branch shows one of the two commits patch-equivalent to something already on trunk and one still pending.

Rather than guess, which shape do you want? I can rebase so only the pending commit remains, or rewrite it as a single commit against current trunk, whichever is less work for you to read. Happy to do either, just say the word.

jackspace:
Correction to my note above, it's gone stale. The branch is a single commit now and none of it has landed, so there's nothing half applied here and nothing for you to pick between.

`git cherry trunk fix/acceptance-doc-counts` gives `+ f77288cf`, the one commit, still pending. Sorry for the noise.

jackspace:
Rebased onto current trunk. 714b0ae4 changed the arithmetic underneath this one: the T-Echo now owes results for both S140 foundations, so the row count and the roster assignment count are different numbers for the first time, twelve rows filled from ten assignments. The check now derives both instead of assuming rows = boards x surfaces. It reads each board's transport from `release/flash/boards.json` and fails closed if that file ever disagrees with `SHIPPING_BOARDS`. Re-proved it in both directions on the new tree: red on a drifted count, red on an unregistered count, red on a mismatched board set, green as committed.

KenAKAFrosty:
Good stuff, nice to clean up the drift itself but also help prevent it in the future

PR #68, opened 2026-08-05
flash: name Firefox in the Web Serial guidance now that it supports it

jackspace:
Follow-up to #67, where the measurements behind this live.

Firefox 151 shipped the Web Serial API on Desktop, enabled by default. The capability probe in `docs/website/src/pages/flash/bridge.rs` is feature detection, not brand sniffing, so it already admits Firefox correctly:

```js
if (!(window.isSecureContext && navigator.serial && navigator.serial.requestPort)) {
  return "unavailable";
}
```

Nothing in the gate needs to change. The guidance copy is what went stale, and it turned out to live in three places, so this updates all of them plus the specs that pinned the old wording:

| where | what it is |
|---|---|
| `src/pages/flash/model.rs` | the `WebSerialCapability` explanations rendered into `#flash-status` |
| `src/pages/flash/view.rs` | the install-message block rendered above it |
| `web-flasher/src/core.js` | `unsupported_browser` in the JavaScript bridge |

Changing only one of them would have left the page telling a Firefox user two different things at once, which is worse than the original staleness.

The three spec assertions no longer pin a browser list. They match the stable part of each sentence, so the next browser to gain Web Serial will not break the suite.

The desktop qualifier in the new wording is load-bearing: Firefox for Android does **not** expose `navigator.serial`. I measured that on Firefox 153.0.3 on Android 16, served over `adb reverse` so the page was a `localhost` secure context and a false negative from an insecure origin was ruled out.

| browser | version | `isSecureContext` | `navigator.serial` |
|---|---|---|---|
| Firefox Desktop, Windows 11 | 153.0.3 | yes | yes |
| Firefox Android, Android 16 | 153.0.3 | yes | no |
| Chrome Android, Android 16 | 150 | yes | yes |

Ran on Windows 11, `x86_64-pc-windows-msvc`:

- `npm run test:browser` - **30 passed**, including the three rewritten assertions
- `npm run test:flasher` - 60 passed
- `cargo test --manifest-path docs/website/Cargo.toml --locked` - 36 passed, 1 failed. The failure is `pages::benchmarks::tests::rewrites_results_links_to_site_routes`, unrelated to this change and reproducible on an unmodified worktree of trunk. It is a line-ending bug I will report separately.

One note in case it is useful: running `npm run test:browser` on Windows needs the signed browser fixture checked out with LF endings. Nothing pins it, so a default Windows clone rewrites it and `build-fixture.mjs` rejects it before Playwright starts. Separate from this PR, and I will raise it on its own.

### DIFF
--- a/docs/website/src/pages/flash/model.rs
+++ b/docs/website/src/pages/flash/model.rs
@@ -59,10 +59,10 @@ impl WebSerialCapability {
         match self {
             Self::Checking | Self::Supported => None,
             Self::AndroidBluetoothOnly => Some(
-                "Web Serial on this Android browser reaches Bluetooth serial devices only, so a USB-connected board never appears in the port picker. Use desktop Chrome or Edge, or the standalone CLI.",
+                "Web Serial on this Android browser reaches Bluetooth serial devices only, so a USB-connected board never appears in the port picker. Use desktop Chrome, Edge, or Firefox 151 or later, or the standalone CLI.",
             ),
             Self::Unavailable => Some(
-                "Web Serial is unavailable in this browser or context. Open this page in current Chrome or Edge over HTTPS, or use the standalone CLI.",
+                "Web Serial is unavailable in this browser or context. Open this page over HTTPS in current desktop Chrome, Edge, or Firefox 151 or later, or use the standalone CLI.",
             ),
         }
     }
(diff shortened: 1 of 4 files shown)


KenAKAFrosty:
Well this was a pleasant bonus! Thanks; it was applied in c6a1673a, and 65f441c2

PR #69, opened 2026-08-05
fixtures: keep the signed browser fixture byte-exact on every platform

jackspace:
`npm run test:browser` cannot run on Windows. Not "fails a test" — it dies before Playwright starts, and I think the cause is worth fixing at the root rather than in the script.

The fixture under `docs/website/web-flasher/browser/fixtures/signed-candidate` is byte-exact by design. `build-fixture.mjs` parses a minisign key ID from a fixed line, requires the channel document's `manifest_sha256` to equal the manifest bytes, and checks every artifact's `size` and `sha256`.

Nothing pins its line endings:

```
git check-attr text -- <any fixture file>
  text: unspecified
```

So with `core.autocrlf=true`, which is the default on a Windows install, all 15 files are rewritten on checkout and none of those checks can pass:

```
claimed        : df6f429fa9ec7555be420647ad039d7d62808bf76fb71daa92a073348dccc423
as checked out : 24f56b8421e0cfda364e3ff50effe8c106ea8c0ec021acd3084731640567b51a  <- mismatch
LF-normalised  : df6f429fa9ec7555be420647ad039d7d62808bf76fb71daa92a073348dccc423  <- match
```

The first failure is earlier and less obvious than the hash: the key ID line arrives with a trailing `\r`, the `$`-anchored regex cannot match, and the run stops at *"the browser fixture public key has no canonical key ID"*.

I started patching that regex, then realised it was the wrong fix. Reverting the patch and normalising only the fixture proves it:

```
staged signed browser fixture 0.2.6 (de81281d26467f670d61d80bb66f1484597267686ab67b4a5eb438310b11aa59)
30 passed (56.9s)
```

The whole lane passes on Windows with **no source change at all**. One `.gitattributes` rule repairs the key-ID parse, the manifest hash, and every artifact hash together, which patching the parser would not have done. It matches how `*.mu` is already handled.

One practical note: this fixes fresh clones. An existing working tree keeps whatever it already has until those paths are re-checked-out or renormalised.

Ran on Windows 11, `x86_64-pc-windows-msvc`, with `dioxus-cli` 0.7.5 and the pinned Playwright 1.61.1:

- `npm run test:browser` - **30 passed**, with `.gitattributes` as the only change in the tree

Found while setting this lane up locally for the first time, to verify #68.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,11 @@
 *.mu -text
 *.sh text eol=lf
+
+# The signed browser fixture is byte-exact: a minisign key ID parsed from a fixed
+# line, a channel document whose manifest_sha256 must equal the manifest bytes, and
+# per-artifact size and sha256 checks. Line-ending translation rewrites those bytes
+# and invalidates every one of those checks, so keep it verbatim on every platform.
+docs/website/web-flasher/browser/fixtures/signed-candidate/** -text
 .githooks/* text eol=lf
 tools/prns text eol=lf
 tools/prns.cmd text eol=crlf


KenAKAFrosty:
The cursed windows line endings strike again!  I've got a windows machine on my desk here, and have ran the prnsd/hopspot/"consumer" side on it quite a bit, but admittedly have spent most of my dev-side efforts on Mac & Linux. Thanks for looking out and filling in the gap I left!  Was rolled into 293fc81f

jackspace:
Ha, they really do. Glad it helped.

Windows is the one platform I can hammer on without it costing you anything, so I'll keep pointing things at it. While chasing something else tonight I ran `tools/tests` on Windows against current trunk and got 26 failures out of 199. A good chunk look like the same family: scripts invoked through their shebang, and `bash` resolving to the WSL launcher instead of the Git one. I've got a branch that takes it from 26 to 13, which is what #71 is, though it needs your call on what the suite should promise on Windows before it's worth much.

No rush on any of it. Just flagging that the gap is bigger than the line endings.

PR #73, opened 2026-08-06
nnpages: dispatch each control request once

jackspace:
`next_refresh_request` picks the lowest-id pending request and returns it, but it never claims it.
The request file is only removed at the end of `finish()`, which runs after the spawned work
completes. Meanwhile the daemon's select loop re-enters that poll as soon as it has spawned the
task, and the poll fires again every `CONTROL_POLL_INTERVAL` (100ms). So any refresh that takes
longer than 100ms is still sitting there on the next pass, and it gets dispatched a second time
under the same id.

I found this chasing an intermittent failure in
`nnpages_live::foreground_daemon_reconciles_page_paths_on_operator_request`:

```
Refreshed 2 hosted route(s): 0 added, 0 removed, 2 unchanged.
```

The count is right and only the delta is wrong, which is the tell. A stale count would mean the
daemon hadn't noticed the deletion. A correct count with an empty delta means something else had
already reconciled it, so the operator's own refresh legitimately had nothing left to report. That
something else is the duplicate dispatch of the previous request, still running, picking up the
deletion the test made in between.

Reproduction on Ubuntu 24.04, the same image CI uses:

```
unmodified trunk                       25 runs, 0 failures
250ms delay added to the spawned work  20 runs, 20 failures, all identical:
  nnpages_live.rs:225 | Refreshed 2 hosted route(s): 0 added, 0 removed, 2 unchanged.
```

Widening the window past the poll interval turns it from a flake into a certainty, and the message
matches the CI failure exactly.

The same undeduplicated path carries announces, so this gap also means one `prnsd nnpages announce`
can put two announces on the air.

The fix claims the request by consuming its file at dequeue. The unlink is atomic, so exactly one
reader wins and a request cannot be handed out after it has been taken. I went with consuming
rather than renaming to a `claimed-` prefix on purpose: renaming leaves a new class of file in the
control directory that nothing ever reclaims if the daemon dies mid-refresh, and that directory
already collects orphaned `result-*` and `*.tmp-*` files on a hard exit. Consuming adds nothing to
clean up. It also lets `PendingNnPagesRefresh` drop its request path, since by the time `finish`
writes the result the file is already gone.

One behaviour change worth naming: if the daemon is killed mid-refresh, the request is gone rather
than retried on the next start. The CLI gives up after `CONTROL_TIMEOUT` anyway, and the reconcile
is idempotent and re-runs at boot and every `NNPAGES_REFRESH_INTERVAL`, so I don't think there's
anything worth preserving there. Happy to keep the file and sweep it at startup instead if you'd
rather this stayed at-least-once.

The new test fails against the current dequeue and passes with the fix:

```
(claim reverted, test kept)
test nnpages::tests::a_control_request_is_dispatched_only_once ... FAILED
  an in-flight request was dispatched a second time

(with the fix)
test nnpages::tests::a_control_request_is_dispatched_only_once ... ok
```

Ran on Ubuntu 24.04 under WSL against trunk `98820b6f`: `cargo test` from `prnsd/` green at
181 + 7 + 1 + 12 + 3 + 3 + 4,
`cargo clippy --all-targets -- -D warnings` clean, and the widened build 20/20 green once the fix
was in. `bash validation/hygiene/fmt-docs.sh`, `python validation/run.py verify` and
`./tools/prns verify` green on Windows. I did not run the full validation suites.


KenAKAFrosty:
Great find. I took the spirit of this forward and include some hardening around the same conceptual issue. Thanks!

PR #77, opened 2026-08-07
Merge trunk into main

KenAKAFrosty:
Brings main current with trunk so the default branch reflects the latest work. Conflicts between the 0.3.3 promote snapshot and newer trunk resolved in trunk's favor; release-gate content from #60/#62/#64 verified present. Not a release action.


KenAKAFrosty:
This is stale now and there's no pressing need to complete it. Shortly we'll have a fresh update to main and 0.3.4 candidate; better to just cancel this

PR #79, opened 2026-08-07
validation: stop treating vendored crates as first-party in the unsafe audit

jackspace:
The deny lane fails on trunk since 27c72282 vendored esp-radio 0.18.0 under a [patch.crates-io] path override. Cargo reports source: null for path dependencies, the same shape as our own crates, so assert_first_party_policy swept the vendored crate into the first-party set and demanded #![forbid(unsafe_code)] on a crate with about a thousand unsafe sites across 63 files. The UNSAFE_EXCEPTIONS path doesn't work either: it then requires denying undocumented_unsafe_blocks across all of them.

Vendoring a dependency doesn't make us its author. The audit now skips the first-party policy assertions for any in-repo manifest under a vendor/ directory. Two exist now — esp-radio 0.18.0 and, since 558c6217, esp-phy 0.2.0 — and the regenerated baseline covers both. Their unsafe still counts into the snapshot, so the surface stays visible: the baseline records esp-radio at 1,012 unsafe tokens and esp-phy alongside it, each under its vendor path as the source.

Clearing that hard failure exposed a second latent one: the committed unsafe-snapshot.json predates the vendoring, so the audit then fails on snapshot drift. Regenerated via --write as the tool instructs. The baseline diff also picks up unrelated dependency-graph changes from recent trunk work (entries leaving the embedded graphs, proto-ipv4 feature additions) that were masked behind the hard failure.

Since the overnight bt-hci git pin, the deny job dies even earlier, at cargo-deny's sources check, before this audit code runs. With a one-line allow-git entry for that pin applied first, the lane advances exactly to the failure this PR fixes — `unsafe audit failed: shipped first-party target lacks forbid(unsafe_code): personal-hopspot/embedded/esp32/vendor/esp-phy-0.2.0/src/lib.rs` — and with both applied together the deny job runs green end to end on a fork run of ci.yml against current trunk: https://github.com/jackspace/Prns/actions/runs/31272372199

Verified: python validation/security/unsafe-audit.py exits 0 on this branch. Run on Windows, Python 3.13.

### DIFF
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -2361,7 +2361,7 @@ SOFTWARE.
 
 License: MIT License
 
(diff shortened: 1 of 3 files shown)


KenAKAFrosty:
Sorry about this; trunk's in a bit of an amorphous state as I'm doing final cleanup on the embedded side for BLE and some wifi stuff. Good clean change here, very welcome. Thanks 🫡

PR #81, opened 2026-08-07
release: keep the pinned Minisign public key byte-exact on Windows checkouts

jackspace:
The published 0.3.3 Windows CLI can't verify any signed release. Every flash attempt fails at validating_manifest with "the Minisign signature is invalid: Invalid encoding in minisign data", error_code release_trust. I hit this today running the Windows acceptance rows against the signed candidate.

Root cause: PINNED_MINISIGN_PUBLIC_KEY is include_str!("../../release/keys/minisign.pub"), so every release binary carries whatever bytes its build checkout produced. Nothing in .gitattributes covered that path, and the Windows CLI is the only target built on a Windows runner, where the default core.autocrlf rewrites the key to CRLF at checkout. Reading the shipped hopspot-flash.exe confirms it: the embedded key at offset 4902208 contains CRLF line endings. The release's own minisign.pub asset is LF and verifies the same manifest and signature fine with standalone minisign on the same host, so the key material is right and only the embedded copy is mangled.

The fix pins the key path to LF in .gitattributes so every checkout produces identical bytes regardless of platform autocrlf settings.

Practical impact until this lands in a release: hopspot-flash on Windows is unusable for its core purpose, while the web flasher works fine on the same host since its verification ships in the website bundle built on a Linux runner.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 *.mu -text
+*.pub -text
 *.sh text eol=lf
 
 docs/website/web-flasher/browser/fixtures/signed-candidate/** -text
--- a/prns-flash-manifest/src/trust.rs
+++ b/prns-flash-manifest/src/trust.rs
@@ -101,6 +101,17 @@ mod tests {
     }
 
     #[test]
+    fn pinned_public_key_is_stored_with_line_feed_endings() {
+        assert!(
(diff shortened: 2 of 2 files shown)


KenAKAFrosty:
Getting tons of Windows dev/admin side cleanup done today. All your findings have been very very helpful. Thank you!

PR #95, opened 2026-08-10
WIP: Heltec V3 compact radio profile

bquilty25:
WIP, not ready to merge.

Developed with GPT-5.6 Terra in GitHub Copilot.

This is a checkpoint for getting the Heltec WiFi LoRa 32 V3 running within its internal RAM limits.

What works on hardware:
- flashing and serial logging
- OLED
- USB runtime
- LoRa, including persistent EU869 configuration
- restoring persisted state

This branch also adds a reduced Wi-Fi station profile and removes the V3 AP / captive-portal services.

Wi-Fi is not usable yet. The smallest valid ESP radio configuration still runs out of internal memory, while smaller buffer settings are rejected by the driver. BLE has the same general memory problem.

The next step is a smaller V3-specific networking path rather than trying to fit the existing Wi-Fi Auto runtime onto the board.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "rtnode-heltecv4"]
+	path = references/rtnode-heltecv4
+	url = https://github.com/jrl290/RTNode-HeltecV4.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "rtnode-heltecv4"]
-	path = references/rtnode-heltecv4
-	url = https://github.com/jrl290/RTNode-HeltecV4.git
--- a/personal-hopspot/README.md
+++ b/personal-hopspot/README.md
@@ -50,6 +50,7 @@ statically, which requires CMake and the Visual Studio Build Tools C++ workload
 
 ESP32 firmware, from `embedded/esp32/` with the board on USB:
 
+    cargo heltec-v3-flash
     cargo heltec-v4-flash
     cargo heltec-v4-r8-flash
     cargo tbeam-supreme-flash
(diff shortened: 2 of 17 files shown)


jackspace:
Nice to see someone else pushing on making the S3 tree fit a smaller board. I have been in these same files this week from the other end, so a few things that might save you a rebase headache.

The big one: this is based on `main`, and the work lands on `trunk`. Right now trunk is 253 commits ahead of main, so whatever happens here is going to need a rebase before it can go in. Worth doing sooner than later, because trunk has been moving fast.

Related, and it should actually shrink your diff: the SX1262 is already optional on trunk as of today, so the S3 tree no longer requires `lora` at all. Your `lib.rs` change and that one are touching the same `compile_error!`, and once you rebase I think a good chunk of your version just falls away rather than conflicting.

The other thing you might find useful is the pattern the tree already uses for a radio that is not there. Rather than branching the code on feature combinations, each radio gets handed to the shared code as `Option<InterfaceId>` and `Option<&'static EmbassyInterfaceStatus>`, declared on both the `#[cfg]` and `#[cfg(not(...))]` arms with the same name and type. The HaLow block in `firmware.rs` is the cleanest example. Because of that, `build_snapshots`, `build_cards` and `classify_card` in `display.rs` have no `#[cfg]` in them at all, and adding another optional radio does not multiply the branches. With four radios in play the combinatorial version gets painful quickly, and I say that having just done exactly this for a board with no 2.4 GHz radios at all.

Happy to be a second pair of eyes on the compact profile once it is on trunk. I have a bench with several S3 boards on it, so if there is something you want run on real hardware, say the word.

bquilty25:
Thanks for calling out the ongoing work on `trunk`, that was very helpful. I had not appreciated how much had moved there already. I have rebased this WIP onto `trunk` and will keep the compact V3 work aligned with the optional-radio patterns there.

evelant:
@bquilty25 I'm curious, what sort of problems did you run into getting things to run on an esp32s3 without psram? I have a couple Heltec Wireless Tracker V2 boards that I purchased without realizing that they were esp32-s3fn8 (same as heltec lora v3) which does not have any psram. It would be nice to get those boards running, but the ram constraints are really tough.

PR #96, opened 2026-08-10
hopspot: make the SX1262 optional in the S3 firmware tree

jackspace:
Every S3 board in the tree so far has carried an SX1262, so the shared S3 firmware requires the `lora` feature outright: `lib.rs` hard-errors without it, and the LoRa interface id, status, and spectrum plumbing run unconditionally through the render loop, the sleep and wake paths, and the card toggles. A board without that radio can't compile the tree today, and boards like that are now real: I've got a LilyGO T-Halow on the bench (S3 plus a Wi-Fi HaLow module, no SX1262), and the direction of #95 is the same tree stripped down for boards that can't afford the full feature set.

This makes `lora` a real per-board choice, the same way `esp-now` already is, and by the same mechanism: Option shims. The card, toggle, sleep, and menu paths see `None` instead of forking the render loop, the lane count follows the feature set (`3 + lora + bluetooth-auto + esp-now`), and the SX1262 board modules plus the Heltec RX-gain constants gate on the feature so a radio-less board never has to invent hardware it doesn't carry. Boards with the radio select `lora` exactly as they did before and compile unchanged.

What I ran, and where:

- `bash validation/hygiene/fmt-docs.sh` green (all three markers), both on my Windows bench and on an ubuntu-24.04 box at CI's pinned toolchain.
- `cargo heltec-v4-r8` (release, Xtensa, Windows esp toolchain): exit 0, zero warnings. The existing lora shape is the same feature set as before.
- The new no-lora shape type-checks on Xtensa (`cargo check --no-default-features --features bluetooth-auto,wifi-auto,esp-now,usb,tcp`). No board in this PR selects it, so that check is synthetic and its dead-code warnings are expected; on my HaLow branch a real board package built on this exact mechanism compiles release with zero warnings and boots on hardware.
- Host side: `personal-hopspot-core` tests (128 passed) and clippy clean on Linux.

No dependency, manifest, or release-graph changes: this is `src/` only. The point of landing it on its own is single-source-of-truth for the S3 tree. My T-Halow board package wants it, and the compact V3 profile in #95 is pushing the same files the same direction, so the shared mechanism seems worth having in first rather than each board growing its own forks.

### DIFF
--- a/personal-hopspot/embedded/esp32/src/lib.rs
+++ b/personal-hopspot/embedded/esp32/src/lib.rs
@@ -8,14 +8,13 @@ extern crate alloc;
     not(all(
         feature = "bluetooth-auto",
         feature = "esp-now",
-        feature = "lora",
         feature = "tcp",
         feature = "usb",
         feature = "wifi-auto"
     ))
 ))]
 compile_error!(
-    "ESP32-S3 firmware is built through a board package, which selects bluetooth-auto, esp-now, lora, tcp, usb, and wifi-auto"
+    "ESP32-S3 firmware is built through a board package, which selects bluetooth-auto, esp-now, tcp, usb, and wifi-auto (plus lora on boards with an SX1262)"
 );
 
 #[cfg(all(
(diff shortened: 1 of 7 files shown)


KenAKAFrosty:
Love this. A very welcome piece of cleanup. Thank you for doing this!

PR #100, opened 2026-08-11
fix: keep the brand splash inside the 64px panel

jackspace:
I flashed a Heltec V4-R8 with the published 0.3.5 candidate today and the boot splash comes up reading "Personal H" and then stops halfway through the "o" of "Hopspot".

The panel is 64px wide (`WIDTH` in `screen/render/layout.rs`). `splash()` draws its text in `FONT_6X10` starting at x=2, so ten characters fit before the edge. "Personal Hopspot" is sixteen characters, which wants 96px and overruns by 34px. `embedded_graphics` `Text` doesn't wrap, so it just draws past the edge and the `DrawTarget` clips it partway through the eleventh glyph. That eleventh glyph is the "o", which is exactly what's on the screen.

"starting" is 50px and "connecting" is 62px, so both of those fit. The brand was the only splash content affected, which is probably why it went unnoticed.

The fix is to split the brand onto the two lines it already is. A portrait 64x128 panel has the vertical room and the whole name survives.

What I think actually matters here is why it shipped: `splash()` had no test. `UiNotice` already has the pattern for this exact problem, a `cfg(test)` `ALL`, a `lines()` accessor, and `every_notice_line_fits_its_rendered_font`, so I gave `SplashContent` the same shape rather than inventing anything. The new test asserts every splash line fits the font it's drawn in, and it fails on the old single-line brand with:

```
Brand line "Personal Hopspot" is 16 characters, more than the 10 that fit
```

I also kept a second small test asserting the brand still says the whole name, so nobody "fixes" a future overflow by quietly dropping a word.

Found while running the 0.3.5 physical acceptance rows on a V4-R8 (Windows host, both the CLI and web surfaces). It reproduced identically after each, so it isn't an artifact of how the board was flashed. Happy to attach a photo of the panel if that's useful.

### What I ran

- `cargo test -p personal-hopspot-core` - 130 pass
- `cargo clippy -p personal-hopspot-core --all-targets -- -D warnings` - clean

Tests and clippy on WSL Ubuntu, gates in Git Bash on Windows. I haven't run this on a T-Echo or any e-ink panel, so I can't speak for how it looks there.


KenAKAFrosty:
Nice lil bit of cleanup here too thanks

PR #103, opened 2026-08-11
interfaces: frame-level RX accounting, carried to the USB peer behind a capability bit

jackspace:
This one comes out of a long bench night. Two boards on a radio link, a daemon on each side holding the board's only USB port, and forty addressed messages that never arrived while every announce crossed fine. The question that mattered was simple: are the frames dying on the air, or arriving and being thrown away? And there was no way to answer it, because `rx_bytes` moves either way. A frame that arrives and fails reassembly has already been counted by the time it's discarded, so from outside the node a link dropping everything is byte-for-byte indistinguishable from a healthy idle one. Taking the console to look kills the daemon that keeps the link alive, and with it the traffic you're trying to observe.

Two changes, one per commit.

**Frame accounting.** `FrameAccounting { frames_in, frames_out, malformed, undecodable, delivered }` joins `InterfaceStatus` as a defaulted method returning `Option`. `frames_out` counts what the medium accepted for transmission, which is deliberately not the claim that RF left the antenna - a link can also die by transmitting into a void, and the TX/RX split is what gives that state a signature. `None` means the family doesn't count, and a caller must not read that as all-zero: "unaccounted" and "nothing arrived" are different answers, which is the whole point. It's defaulted, so no existing family, binding, or FFI surface changes. `EmbassyInterfaceStatus` backs it with atomics behind an opt-in `account_frames()` declaration, so a family that never declares stays `None` even if a stray increment lands, and a test pins that.

**Carrying it to the peer.** `Message::Vitals` (kind 0x04) in the usb_auto protocol carries one interface's id, connection state, the node's own monotonic uptime (so every sample dates itself even across host polling gaps - when the question is "when did arrivals stop", host-side timestamps fold polling gaps into the timeline), byte counters, and the optional accounting block, with a presence flag so `None` survives the wire as `None`. It sits behind a new `Capabilities` bit: a sender only emits it toward a peer that advertised the bit, so a pre-vitals peer never sees the kind byte at all, and lane negotiation ignores the bit entirely (the composition test covers both directions, and your existing "unknown capability bit is preserved, not rejected" test already had the forward path covered, which is a nice piece of design). The device reaction ignores an inbound report; the host reaction surfaces it as `HostInbound::Vitals`.

Deliberately not in this PR: nothing sends the message yet and the host does nothing with it past parsing. The sender needs a per-board list of statuses and a cadence decision, and the host needs a store and somewhere to expose it. Both are worth their own conversation once this seam exists. What's here is complete at its own layer: codec, negotiation, both reaction surfaces, and the accounting source it carries.

Testing, all host-side on Linux: the usb_auto suite (30 tests) including round-trips with and without the accounting block, truncation rejected at every cut point, a presence flag that contradicts the length rejected, and the three existing property tests now generating arbitrary `VitalsReport`s through the payload and framed-stream codecs. The accounting tests were proven red first: with the increments stripped, the counter tests fail and only the overcount guards pass. `cargo test` and `clippy -D warnings` green on prns-core, prns-runtime-embassy, prns-interfaces-tokio, and prns-interfaces-embassy, plus the workspace check, the fmt/docs gate, and the validation registry; the test-and-clippy pass repeated at 1.96.0 to match the pinned CI toolchain. One small enabling change: `ConnectionState::as_u8`/`from_u8` lose their host-feature gate because the ungated codec now needs them.

### DIFF
--- a/personal-hopspot/mobile/ios/rust/Cargo.lock
+++ b/personal-hopspot/mobile/ios/rust/Cargo.lock
@@ -1327,6 +1327,7 @@ dependencies = [
  "embedded-storage-async",
  "heapless 0.8.0",
  "personal-rns",
+ "prns-macros",
  "sha2 0.11.0",
 ]
 
@@ -1430,6 +1431,7 @@ dependencies = [
  "heapless 0.8.0",
  "hkdf",
  "hmac",
+ "prns-macros",
  "rayon",
  "roaring",
  "sha2 0.11.0",
@@ -1486,6 +1488,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "prns-macros"
+version = "0.3.6"
+
+[[package]]
 name = "prns-runtime"
 version = "0.3.6"
 dependencies = [
@@ -1493,6 +1499,7 @@ dependencies = [
  "cfg-if",
  "heapless 0.8.0",
  "prns-core",
+ "prns-macros",
 ]
 
 [[package]]
--- a/personal-hopspot/mobile/ios/rust/src/usbmux.rs
+++ b/personal-hopspot/mobile/ios/rust/src/usbmux.rs
@@ -162,7 +162,10 @@ async fn serve_stream<Seam: InterfaceSeam>(
                         Ok(Message::Data(packet)) if !packet.is_empty() => {
                             seam.next_inbound(packet).await;
                         }
-                        Ok(Message::Data(_)) | Ok(Message::HelloAck { .. }) | Err(_) => {}
+                        Ok(Message::Data(_))
+                        | Ok(Message::HelloAck { .. })
+                        | Ok(Message::Vitals(_))
+                        | Err(_) => {}
                     }
                 }
             }
--- a/prns-core/src/interfaces/mod.rs
+++ b/prns-core/src/interfaces/mod.rs
@@ -50,8 +50,8 @@ pub use policy::{
     TransportCapability, LOCAL_INTERFACE_BITRATE_ESTIMATE, TRAVERSED_NETWORK_BITRATE_ESTIMATE,
 };
 pub use status::{
-    AirtimeUtilization, ConnectionState, InterfaceSnapshot, InterfaceStatus, InterfaceVitals,
-    Membership, TransferRates,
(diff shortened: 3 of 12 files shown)


jackspace:
Added one more field while this sits unreviewed and the wire is still free to change: `last_frame_in_at_ms`, the reporter's clock when the interface last accepted an inbound frame. The five accounting counters are monotonic totals, so a single sample can't tell a frozen interface from a quiet one, and I got to watch that failure mode for real this week: a wedged radio module passed zero frames for hours while every counter a host could poll read healthy. Age falls out of `uptime_ms - last_frame_in_at_ms` in one report, no second sample, no host clock. `u64::MAX` spends an unreachable value on "no frame since boot" so never doesn't collapse into a number.

KenAKAFrosty:
Thank you for this! The need behind it is extremely valid, and the concrete failure case is valuable too. We definitely need to distinguish radio silence from frames arriving and dying during decoding or reassembly, without disrupting the device to attach a console.

I think we should hold off on merging this though, beacuse I don’t quite think we should establish a USB-specific Vitals schema here.

We already have InterfaceVitals, InterfaceSnapshot, local status, observability, and RNS remote-management reporting. We are also about to deepen Remote Management substantially. I’d like this requirement to shape that unified model so the same authoritative interface report can be exposed over local RPC, Reticulum Remote Management, USB, or any other transport without parallel schemas or conflicting sampling semantics.

I’m going to keep this PR open for now to be treated as requirements for Remote Management, which such valuable things as: 
- Unavailable versus zero accounting
- Cumulative receive-path outcomes
- Recent useful activity measured on the device
- and backward-compatible capability/version negotiation

The implementation would then live in the common interface-reporting model rather than in a USB-only message.

So this is not a rejection of the need (quite the opposite!). I think the contribution has identified an important prerequisite for the management work. My preference is to extract the typed reassembly improvement, then supersede the transport and reporting portions with the unified Remote Management that will begin extremely soon.


Thanks again!
Ken

tonycowan:
Thanks for this work @jackspace ! I'm running into similar needs while debugging. Looking forward to the new management UI to provide some visibility then we can expand what there is to view.

KenAKAFrosty:
Hey y'all, as we've all talked about, this will be handled in the larger sense by #131 , so just gonna close this out now

PR #104, opened 2026-08-11
fix Heltec V4-R8 first-render I2C panic from private PSRAM clobber

tonycowan:
## Summary
- Heltec V4-R8 panicked on the first post-radio UI flush with an I2C divide-by-zero in `esp-hal` (`Config.frequency` became 0).
- Root cause: `run_core` was pinned into the R8 private PSRAM bump (`PsramAlloc`), then `reinit_private_psram_heap()` reset that bump before allocating the LoRa TX queue — overwriting the live future (including OLED/I2C state). Heltec V4 (R2) was unaffected because `PsramAlloc` there is global `ExternalMemory` and reinit is a no-op.
- Fix: pin `run_core` into `esp_alloc::ExternalMemory` instead of `PsramAlloc`.

## Test plan
- [x] Smoke Heltec V4-R8 with this build: `first-render.complete`, no panic (MAC `ac:a7:04:e1:4b:3c`)
- [x] Confirm unfixed trunk still boots cleanly on Heltec V4 R2 (`first-render.complete`)
- [ ] Optional: re-smoke R2 with this branch for shared-path confidence

### DIFF
--- a/personal-hopspot/embedded/esp32/src/s3/firmware.rs
+++ b/personal-hopspot/embedded/esp32/src/s3/firmware.rs
@@ -4,7 +4,11 @@ pub(crate) async fn run<B: Esp32S3Board>(spawner: Spawner) {
     let config = esp_hal::Config::default().with_cpu_clock(CpuClock::max());
     let p = esp_hal::init(config);
     let bringup = B::bringup(p).await;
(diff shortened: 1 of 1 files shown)


KenAKAFrosty:
Welcome back! Thanks a ton for getting this smoked out on your R8, and for the fix

PR #111, opened 2026-08-13
Add Heltec Mesh Node T114 Hopspot support

mark-ik:
## Summary

- add a Heltec Mesh Node T114 Rev. 2.x target to the nRF52840 Personal Hopspot firmware
- keep the board-specific clock, SX1262 wiring, RF switch, flash layout, storage, identity, and USB startup behind the T114 board module
- fix shared SX126x receive re-entry after `RxDone` and `TxDone`
- record the exact current-main USB and bidirectional RF hardware qualification

## Why

The T114 combines the nRF52840 and SX1262 already represented in the embedded stack, but it needs its own board authority for pins, external crystal startup, TCXO and DIO2 RF-switch control, storage ranges, and UF2 memory layout.

The hardware pass also exposed a shared SX126x lifecycle defect. Re-entering receive immediately after an IRQ could race a still-high DIO1 line, and a completed receive needed an explicit re-arm. Keeping those corrections in the shared radio driver avoids a T114-only workaround.

## Impact

Developers can build `heltec-t114` with the `board-t114` feature and convert the resulting binary to an nRF52840 UF2 at application base `0x26000`. The image exposes LoRa plus USB Auto as `Personal Hopspot (Heltec T114)` / `PERSONAL-RNS-T114-HOP`.

The public signed-flasher catalog does not carry the T114 yet. This PR adds the developer target and its verified hardware boundary; it does not promote a public recovery package.

## Validation

- `cargo build --release --locked --no-default-features --features board-t114 --bin heltec-t114`
- `cargo build --release --locked --no-default-features --features board-t-echo --bin t-echo`
- `cargo check --locked --no-default-features --features board-t096 --lib --target thumbv7em-none-eabihf` (the T096 binary retains its intentional scaffold-only stop)
- `cargo test --manifest-path prns-interfaces/impls/embassy/Cargo.toml --features lora --lib` (51 passed)
- `cargo fmt --manifest-path personal-hopspot/embedded/nrf52840/Cargo.toml --all -- --check`
- `git diff --check`
- exact UF2: 694,272 bytes, SHA-256 `f1901becf0b2cb19162180425d5e555a5d67562898743314afc6798333147c2f`
- real T114 USB enumeration and USB Auto `Hello`/`HelloAck` with node tag `t114-usb`
- stock RNS 1.4.2 and an independently flashed Heltec V4 RNode 1.86 at 915 MHz / BW 250 kHz / SF9 / CR4:5:
  - V4 to current-main T114: fresh signed destination crossed RF and arrived over T114 USB
  - current-main T114 to V4: stock RNS validated and dispatched the fresh signed announce with exact application data
- the V4 oracle was restored afterward through the signed Prns Hopspot 0.3.4 package and again completed USB Auto `Hello`/`HelloAck` as `heltecv4`

The receipt proves normal-mode bidirectional single-frame RNode interoperability for the exact current-main T114 image. The older qualification in the same document covers maximum-fragment framing. Physical contention fairness and multi-node queue drainage remain separate gates.


mark-ik:
Follow-up sequencing: this PR intentionally stays limited to the recovery-safe, headless T114 target and its hardware qualification. After it lands, the ST7789 display can follow as a focused PR using the independently verified pin map and HardwareProfile offsets 40,52, while keeping headless as the default. Separately, once the dual-S140 work reaches main, we can consider an opt-in S140 v6 BLE profile without changing the stock bootloader. Either follow-up comes after #111.

jackspace:
Heads up on something easy to miss, because GitHub is cheerfully telling you everything is fine.

This PR targets `main`, and the nRF52840 work all lands on `trunk`. So the "no conflicts" result is real, it is just answering a question nobody asked. Against `trunk`:

```
git merge-tree --write-tree origin/trunk <this branch>
CONFLICT (content): personal-hopspot/embedded/nrf52840/build.rs
CONFLICT (content): personal-hopspot/embedded/nrf52840/Cargo.toml
CONFLICT (content): docs/embedded.md
CONFLICT (content): validation/platforms/embedded.sh
```

The dual-S140 work you mention above already landed on trunk last night, in `25b8aa6b` and `02cc6f01`. It splits `memory.x` into `memory-s140-v6.x` at `0x26000` and `memory-s140-v7.x` at `0x27000`, and picks between them in `build.rs` off the SoftDevice feature. Yours picks off the board feature. Same line, different axis, so they collide rather than stack.

Good news is they compose and it is a small change. I resolved it locally as a board by SoftDevice matrix and built your target on top of current trunk:

```
cargo build --release --locked --no-default-features --features board-t114 --bin heltec-t114

.vector_table   0x00026000
.text           0x00026100
.data           0x20006000
```

Linked exactly where you put it. WSL Ubuntu, rustc 1.97.1, thumbv7em-none-eabihf.

One thing to watch when you rebase: git's rename detection applied your `memory.x` edit to the renamed `memory-s140-v7.x`, so the v7 layout silently turned into your two-line `INCLUDE board-memory.x` stub. It merged without a conflict and it was wrong, which is the combination that actually hurts. Worth a look at that file after the rebase.

On your sequencing note, I have a stock T114 on the bench that has never been re-bootloadered, plus a Windows host, so if an opt-in S140 v6 BLE profile does become a thing after this lands I am happy to put hardware behind it rather than opinions.

Thanks for doing the pin map, crystal startup and RF switch work. That is the part nobody else could have written.

jackspace:
Your msos hoist fixes a Windows bug, and I don't think you knew that when you wrote it.

I have a T114 and a Windows box, and trunk's firmware won't bind a driver there. Device Manager gives Code 28, no COM port, nothing. I pulled the descriptors off the wire from a Linux host to find out why:

```
Set header
  Configuration subset       type 0x0001
    Function subset          type 0x0002
      Compatible ID WINUSB   type 0x0003
      DeviceInterfaceGUIDs   type 0x0004
```

The compatible ID sits inside a function subset. `WebUsbAutoClass` is a single-interface device, so Windows never loads `usbccgp`, so there is no function PDO for that subset to attach to. Windows falls back to the interface class, finds `USB\Class_FF`, has no driver for it, and stops.

Your change is the fix:

```diff
-        let mut function = builder.function(0xff, 0, 0);
-        function.msos_feature(msos::CompatibleIdFeatureDescriptor::new("WINUSB", ""));
+        builder.msos_feature(msos::CompatibleIdFeatureDescriptor::new("WINUSB", ""));
+        let mut function = builder.function(0xff, 0, 0);
```

`function.msos_feature()` auto-opens a configuration subset and a function subset before writing. `builder.msos_feature()` writes at device level, which is exactly what your comment above it says it should be.

Built your branch, flashed it, same board, same machine, same cable:

```
before   Status Error   Code 28   Service (none)
after    Status OK      Code  0   Service WINUSB
         CompatibleIds gained  USB\MS_COMP_WINUSB
```

`USB\MS_COMP_WINUSB` is synthesised by Windows from the MS OS descriptor set, so its appearing is the proof Windows actually read them this time.

Two caveats on what I tested. My T114 has already been re-bootloadered to S140 7.3.0, so I built `board-t114` against a `0x27000` layout rather than your stock `0x26000` one. I verified your fix, not your exact image. And this says nothing about BLE, the SoftDevice is linked but never enabled.

Worth flagging because it widens the PR beyond the T114: `WebUsbAutoClass` is shared, so any board where USB Auto is the only USB function has this same problem on Windows today.

I also spent an hour on a wrong theory first, that Windows had cached a failed install for the VID/PID. Cleared the cache, nothing changed. It was the descriptor placement all along.

KenAKAFrosty:
Thanks again for this!

I’ve now separated & pushed two shared corrections from the T114 port to trunk, with attribution back to #111: The device-level WinUSB descriptor correction, and the SX126x packet-parameter and IRQ-release corrections.

For the radio work, I took the Semtech-supported requirement that every LoRa `SetPacketParams` operation reapply the IQ-polarity workaround, along with waiting for DIO1 to deassert after clearing `TxDone`.

I intentionally didn't take the unconditional `SetRx` after every `RxDone`. We explicitly use `0xFFFFFF` continuous receive mode, which Semtech documents as remaining in RX and automatically searching for the next packet ([in their datasheet](https://semtech.my.salesforce.com/sfc/p/#E0000000JelG/a/RQ000008nKCH/hp2iKwMDKWl34g1D3LBf_zC7TGBRIo2ff5LMnS8r19s) on page 75). 

Our existing evidence with the T-Echo + T-Beam Supreme + Heltec V4 also supports avoiding that repeated reconfiguration. If you have specific insight for why this is still preferred, please share!  The T114 qualification clearly exposed a real stateful problem (very helpful!), but the three changes were tested together, so I would like to isolate that third behavior before making it shared driver policy.

The remaining valuable work here's the T114-specific port itself. Current trunk’s dual-S140 and build-layout work means that portion needs to be reapplied to the current structure rather than merged as-is.

Please leave this PR open as the source (and attribution record) while I carry that board-specific work forward. There's no need to spend time rebasing the full branch right now; I’ll circle back once the trunk-oriented port is ready for comparison

KenAKAFrosty:
Okay! I just landed a focused adaptation on current trunk that brings up the T114 as a headless Hopspot with LoRa and USB Auto while preserving the stock MBR, S140 recovery foundation, and UF2 bootloader.

Would either of you be willing to retry the exact current image on your T114 Rev 2.x?

1. Update to current trunk at c40e8cb0.
2. Run ./tools/prns build hopspot t114.
3. Double-press Reset and wait for the HT-n5262 volume.
4. Copy target/hopspot-t114/heltec-t114.uf2 onto it.

The most useful observations would be:

- Whether it boots and shows heartbeat LED activity.
- Whether USB enumerates as Personal Hopspot (Heltec T114) with VID/PID 1209:0001.
- Whether a PrnsUsbAuto host completes its handshake and exchanges an actual packet.
- Whether LoRa receives and transmits successfully against a known-good peer
- Whether double-reset recovery still mounts the UF2 volume afterward.

Thanks y'all!

jackspace:
Ran it, on a stock board.

Mine was not stock when your comment arrived — I had re-bootloadered it to S140 7.3.0 back on
8 August, which puts the application base at `0x27000`, so your `0x26000` image would have been
written four kilobytes below where the bootloader jumps and done nothing. Rather than report that as
a result, I put the factory bootloader back. `INFO_UF2.TXT` now reads:

    UF2 Bootloader 0.9.0-2-g836c8dc-dirty
    Model: HT-n5262
    Board-ID: HT-n5262
    Date: Jul  9 2024
    SoftDevice: S140 6.1.1

So the below is your image, unmodified, at the layout you intended.

**Boots, heartbeat LED** — yes, confirmed on the stock image with your exact bytes.

**USB enumerates as Personal Hopspot (Heltec T114), 1209:0001** — yes, and cleanly on Windows:

    Status: OK    ConfigManagerErrorCode: 0    Service: WINUSB
    CompatibleIDs: USB\MS_COMP_WINUSB, ...

That is your device-level WinUSB correction holding on hardware. Before it, this board was Code 28.

**Double-reset recovery still mounts the UF2 volume** — yes. Worth adding that after the SoftDevice
downgrade the bootloader found no valid application at `0x26000` and stayed in DFU mode by itself,
so the volume was already mounted without a double-press. The recovery path is doing what it should.

**LoRa receives and transmits against a known-good peer** — yes, both directions. I had to go
outside prnsd to watch it, for the reason in the next check, so I used the browser console you
already ship: `docs/website/public/browser-node-playground-console/`, over WebUSB on Windows. It
found the board and opened a USB Auto session on it, one interface, MTU 8192:

    8:36:22 PM  USB Auto   Device selection opened
    8:36:40 PM  USB Auto   Session opened · Interface 0548e871781dbc56

(that console prints local time; the bench is UTC-6, so those are 02:36:22Z and 02:36:40Z)

That browser node had exactly one interface — no Wi-Fi, no WebSocket — so everything it saw arrived
over the T114's radio and nothing else.

The listener on the other side was an RNode at 915.000 MHz, SF9, 250 kHz, CR 4/5, which is
`DEFAULT_915_PROFILE` from `prns-core/src/interfaces/lora/profile.rs`, the profile
`runtime/headless.rs` compiles in. Its daemon had `enable_transport = No` and no USB Auto interface
at all, so it is a radio and nothing more.

**Receiving.** In three minutes the browser node logged five announces arriving on the T114's
interface at seven and eight hops. What makes them attributable is the RNode: it heard the same five
over the air, in the same second, at exactly one hop fewer.

    destination   heard at       RNode, direct   browser, via the T114
    55b210be…     02:36:41.724Z   6 hops          7 hops
    d24e836d…     02:37:07.035Z   6 hops          7 hops
    29fffb69…     02:37:32.696Z   7 hops          8 hops
    46f755f9…     02:38:52.120Z   7 hops          8 hops
    645ac747…     02:39:16.606Z   6 hops          7 hops

Both sides logged each of those in the same second.

Five for five, no exceptions. That extra hop is the T114 taking the packet off the air and handing
it up the USB link.

**Transmitting.** The browser node's own two destinations then came back the other way, out over the
T114's radio and into the RNode at two hops — the browser is the origin, the T114 is the one relay:

    950d124f3c97ab95871e551aa2cfedd8  nomadnetwork.node  02:37:47.220Z  2 hops
    5d6ee105ae6bda9b725e77e26169275b  lxmf.delivery      02:38:18.183Z  2 hops
    950d124f…                                            02:39:17.507Z  2 hops
    5d6ee105…                                            02:39:41.823Z  2 hops

Twice over: once from the runtime's own startup announce, once from pressing Send LXMF announce.
Nothing else on that bench could have put those two hashes on 915 MHz.

**A PrnsUsbAuto host completes its handshake and exchanges a packet** — yes, as the section above
shows, but not with prnsd, and that gap is worth naming.
`prns-interfaces/impls/tokio/src/usb_auto/host.rs` builds its candidate list in `scan_cdc_targets()`,
which is `scan_usb_serial_ports()` and nothing else, so PrnsUsbAuto attaches over CDC. The T114 in
application mode is `WebUsbAutoClass` — a vendor-class WebUSB interface with no CDC port, which is
exactly why the WinUSB binding mattered. It never becomes a candidate. A fresh 0.3.6 daemon on this
machine attached to a Heltec V4-R8 in 55 ms and carried 161 announces in 25 seconds, so the host
works; it simply cannot see this board.

The part that can see it is already in the tree. `personal-hopspot/desktop/src/host_usb.rs` scans
with `nusb` in `scan_usb_auto_targets()`, and `is_webusb_auto()` matches on `WEBUSB_VENDOR_ID` /
`WEBUSB_PRODUCT_ID` — 0x1209 / 0x0001, which is this board exactly. The browser console under
`docs/website/public/browser-node-playground-console/` reaches it the same way, and that is the one
I used above: session opened, interface up at MTU 8192, packets carried in both directions. So the
check itself passes — a PrnsUsbAuto host does complete the handshake with this board and exchange
traffic. It just cannot be prnsd, because prnsd is the one host that only looks for CDC.

Two smaller notes from the bench.

The radio profile is fixed at 915 and the store is ignored: `runtime/headless.rs` sets
`DEFAULT_915_PROFILE` directly, while `runtime/firmware.rs` loads the store with 915 as a fallback.
That matches what the README says, so it is not a complaint on its own — but paired with the next
paragraph it means a T114 has no way at all to change frequency or region, and `Region::Us915` is
compiled in, which is wrong for most of the world.

`memory-t114.x` reserves two pages for a radio profile store at `RADIO_PROFILE_FLASH_ORIGIN =
0x000E9000`, with an `ASSERT` pinning them against the node identity. `RADIO_PROFILE_PAGES` is
defined only in `boards/t_echo/mod.rs`, and the headless runtime never touches the store. So the
storage is reserved and asserted, and nothing on the T114 path reads it.

That last one feels like it belongs with the remote-management work rather than here — a headless
board with no screen, no BLE and no portal is exactly the case that has to be configured from
somewhere else, and the pages are already carved out for it.

The board is on the bench and stock now, so I am happy to run anything else you want checked.

KenAKAFrosty:
Thanks a ton for testing this and giving that nice thorough report back! Good call on the new web usb stuff, got that work properly unified on  ff1a68c51285b579de84f5483ada73575f797248

If you can confirm that works as expected on that board, that'd be a big help! After that I think we can consider this board properly supported. (I'm also grabbing one to help double-check and refine in the future). 

The main issue is, of course, the remote management limitation for any screenelss board. I'll be opening an issue soon to track that and have it surfaced as a clear high priority

jackspace:
All four confirmed on the T114, Windows, against `ff1a68c5`: WinUSB discovery, reconnect,
recovery-cycle rediscovery, and interface release.

**Setup.** prnsd built from `ff1a68c5` with `cargo build --release --locked`, native MSVC on
Windows 11 (26200), x86_64. Default features, so `tokio-host` pulls `personal-rns/usb` through to
`nusb` without anything extra. Board is a stock Heltec T114, factory bootloader S140 6.1.1, running
your `0x26000` image, enumerating as `1209:0001` with service `WINUSB` and
`ConfigManagerErrorCode` 0. One `PrnsUsbAuto` interface, `enable_transport = No`.

**Discovery works.**

```
usb-auto: opened WebUSB Auto PCIROOT(0)#PCI(0801)#PCI(0004)#USBROOT(0):1 interface=0 in=0x81 out=0x01
interface_connection_changed  previous=Reconnecting connection=Connected
```

17 ms from open to Connected, and it reproduced on five independent starts without a single miss.
The endpoint pair is the same single vendor-class interface the browser host found.

To make sure that was your commit and not the bench, I ran a pre-commit binary against the same
board, the same config, two minutes later: prnsd 0.3.4, from a tree whose `usb_auto` still builds
candidates in `scan_cdc_targets()` and has no reference to `WEBUSB_VENDOR_ID` anywhere. Zero WebUSB
lines. Same hardware, same minute, only the commit different.

**One thing that control turned up, and I think it matters more than the pass.** Both arms reached
`connection=Connected`. The USB Auto host is a single aggregate interface, and the two Heltec V4-R8s
I have on CDC ports attach to it whether or not the T114 did. So on a bench with any other USB Auto
board present, `Connected` isn't evidence the T114 attached. The only thing that distinguishes it
is the `native.rs:501` log line. If someone qualifies this board by looking at interface state, or
at a UI showing the interface up, they get a green that examined nothing. Related, and smaller:
successful CDC opens aren't logged, only failures, so a log can't tell you which ports a running
instance actually holds. It took a second contending instance failing on COM15 and COM16 for me to
establish that the first one had them.

**Interface release works, on process exit.** Two instances, same board. While A held it, B failed
cleanly and backed off rather than storming:

```
19:07:16.488  usb-auto: open webusb:...:1:0 failed: failed to open device (error 5)
19:07:21.498  (same)
19:07:27.492  (same)
```

A exited just after 19:07:27.095, and B claimed the board at 19:07:33.508 on its next retry, with
no restart of B. 6.0 s, one backoff cycle. It's observed from a separate process, so it's release
rather than an assumption about release. The board also stays healthy after prnsd exits, still
Status OK with error code 0.

Worth noting what a leaked handle actually looks like, because I had one running the whole time by
accident: a leftover daemon of mine owned COM4, and every run logged
`open cdc:COM4 failed: Access is denied. (os error 5)` every five seconds. The T114's interface
never produced that signature after an exit.

**Traffic, so this isn't just a handshake.** I put an RNode on the same air as a second, independent
listener at 915.000, SF9, 250 kHz, CR 4/5, with no USB Auto interface on it at all, so anything it hears
came off the air. Over about eight minutes the T114 host took 65 announces up through the native
USB Auto session, and 16 of the 18 destinations the RNode heard also arrived at the host. The
board's receiver is genuinely working and the session carries real traffic.

I can't turn that into a hop claim, though, and I want to be straight about it rather than repeat
the earlier number. Pairing the two listeners by destination and nearest arrival time, only 2 of 18
ear events have a host counterpart inside three seconds: one at +1 hop and one at +0. Two samples
isn't a result. The earlier browser-host run got a clean 5-for-5 at exactly +1 and this run doesn't
reproduce it; I don't have an explanation yet. (Pairing first-occurrence to first-occurrence over
the window makes it look like 15 matched pairs, which is meaningless, because these announces repeat, so
that pairs unrelated events tens of seconds apart.)

**Reconnect works.** Three physical unplug/replug cycles, three rediscoveries, no restart of prnsd:

```
21:43:24.839  usb-auto: opened WebUSB Auto BUS:1 interface=0 in=0x81 out=0x01
21:43:35.851  usb-auto: opened WebUSB Auto BUS:1 interface=0 in=0x81 out=0x01
21:43:45.843  usb-auto: opened WebUSB Auto BUS:1 interface=0 in=0x81 out=0x01
```

11.0 s and 10.0 s between successive attaches, which puts rediscovery under about ten seconds
including the hand doing the replugging.

**The thing I'd most want you to see, though, is what happened next.** The board went off the USB
bus entirely for **3 minutes 28 seconds**. That was my fault, not the board's: I swapped the USB
cable mid-test and the replacement turned out to be charge-only, so nothing enumerated until I put
the original back. I confirmed the absence three times on the Windows side: no `1209:0001`, no
`239A`, a full bus enumeration showing only my other boards. When it came back, prnsd reopened it
with no restart, which is the reconnect answer again on a much longer gap.

The cause doesn't matter for what it exposed, though.

Across that whole 3m28s, prnsd reported nothing:

- zero `interface_connection_changed` events
- zero failed opens on the `webusb:` locator
- the interface sat at `connection=Connected` the entire time
- **52 announces kept arriving on that same interface**, from other boards

The USB Auto host is one aggregate interface, so it stays up on any member. A screenless board can
therefore drop off its host completely, for minutes, while the daemon shows a healthy interface and
a live announce feed. That is the same aggregate problem as the `Connected` note above, but this is
the version with consequences, and it seems worth having in hand for the screenless remote
management issue you mentioned opening.

**The recovery cycle works.** Double-tap into the stock UF2 bootloader, hold it there, single reset
back out:

```
22:02:17Z  1209 gone, drive F: = HT-n5262 mounted
22:02:21Z  239A enumerated, stock UF2 bootloader up
           ... held 2m43s ...
22:05:04Z  F: unmounted (reset pressed)
22:05:05.831Z  usb-auto: opened WebUSB Auto BUS:1 interface=0 in=0x81 out=0x01
22:05:08Z  1209 back in application mode
```

prnsd picked it up within about one to four seconds of the board returning, no restart. The
bootloader reported itself as `HT-n5262`, S140 6.1.1, Jul 9 2024, the stock one, unchanged, after
four insert/remove cycles and a double-tap. I only read `INFO_UF2.TXT`; nothing was written to that
drive.

The session that came back is a working one rather than just an attachment: 184 announces in the
following 714 s, about 15 a minute, the first arriving 17 s after the interface reopened.

That excursion also repeats the silence above, and I think that is the part worth keeping. For the
169 s the board sat in the bootloader, prnsd logged **0** connection changes, **0** failed opens on
the `webusb:` locator, kept the interface at `Connected`, and took **24** announces from the other
boards. Two absences now, 208 s and 169 s, from two unrelated causes, and nothing in the daemon's
observable state moved for either one.

KenAKAFrosty:
Thank you, Mark — the T114 work from this PR is now incorporated and shipping on `main`: the canonical catalog marks `t114` as shipping, the 0.3.7 roster covers both web and CLI surfaces, and the shipping-firmware matrix builds it.

The implementation evolved during the later trunk-to-main promotion, including a runtime/module reorganization, so rebasing this old head onto current `main` now conflicts across renamed and removed paths and would overwrite newer SX126x/T114 code. `main` also requires linear history, which rules out an ancestry-only merge commit.

Your original board-authority documentation and linked GitHub authorship are preserved on the default branch in commit f1ce507c4c61aac4aba8504bec93cf264d476411, and the qualification record links back to #111. Closing this PR as incorporated without changing the current shipping implementation.

PR #113, opened 2026-08-14
release: derive browser fallback scenarios from Web Serial support

jackspace:
Firefox 151 shipped Web Serial in May 2026, and three of the four points every browser fallback row has to prove describe a page that a browser with Web Serial never renders.

`release/acceptance/README.md` requires each row in `browser_fallbacks` to prove all four: the ESP CLI guidance is present, ESP connect is unavailable, no broken connect action is shown, and the T-Echo UF2 route remains available. The first three only exist when the browser cannot reach a serial port at all. On current stable Firefox the flasher takes the supported branch and shows none of them, so `("firefox","macos")`, `("firefox","windows")` and `("firefox","linux")` cannot be completed honestly by anyone. Reproduced on Firefox 153.0.3 on Windows, where the flash itself completed.

The flasher never asks which browser it is in. `web_serial_probe_script` feature-detects `window.isSecureContext && navigator.serial && navigator.serial.requestPort` and branches on the answer. The acceptance contract was the only place pinning a browser list, so this asks the same question the flasher already asks.

`REQUIRED_FALLBACKS` is now the union of `WEB_SERIAL_FALLBACK_BROWSERS` and `WEB_SERIAL_CAPABLE_BROWSERS`, and `fallback_scenarios(browser, os)` returns what that browser can actually observe. A browser without Web Serial still proves all four. A browser with it proves the T-Echo UF2 route, which is a manual copy and does not depend on Web Serial. Same single-source-of-truth shape as deriving the acceptance counts from `SHIPPING_BOARDS` in #52: a browser moves between the two groups by shipping the API, not by being edited here.

Firefox keeps its rows. What a Web Serial capable browser should assert beyond the UF2 route is a release-confidence call and it is yours, so I have left it open rather than deciding it in a patch. The honest options look like keeping a thin fallback row, or moving those browsers onto the web surface where a Web Serial capable browser now arguably belongs. Happy to follow up with whichever you prefer.

The generator and the validator both go through `fallback_scenarios`, so scaffolds and completed rosters cannot disagree.

Tests: a Firefox row claiming `esp-connect-unavailable` is now rejected, and the existing "requires every truthful fallback scenario" test moved to the Safari row, which is the one that still must prove all four. I checked the new test fails against the old flat behaviour rather than assuming it discriminates.

Run on Windows: `python validation/run.py verify` (`VALIDATION_REGISTRY_OK`), `./tools/prns verify` (`TOOLING_REGISTRY_OK`), `bash validation/hygiene/fmt-docs.sh` (`FMT_DOC_CHECK_GATE_OK`), and `tools/tests/test_validate_flasher_acceptance.py`, 30 passed. No Rust changed, so no cargo run.

Follows up #67.


jackspace:
Rebased onto trunk now that #65 is in. The doc contract it added binds the exact sentence this PR rewrote, so I pointed the contract at the new wording ("A browser without Web Serial must prove all four points:") in the same change. Checked it both ways: the gate passes on this tree and still fails if the README sentence drifts.

KenAKAFrosty (changes requested):
The capability premise is correct: current MDN compatibility data records navigator.serial and Serial.requestPort in Firefox 151. The resulting evidence model is not yet release-meaningful, though. All three Firefox entries remain browser_fallbacks and require only t-echo-uf2-route. That route does not exercise the newly supported Web Serial path, and the same manual UF2 observation is repeated on Linux, macOS, and Windows. Meanwhile the acceptance overview still calls these four rows unsupported-browser rows, which is no longer true. Please keep unsupported fallback evidence and supported Web Serial evidence as honestly named concepts. Safari can remain the unsupported fallback. Firefox should either move into the supported physical web coverage or into a distinct capability-derived browser matrix with scenarios that actually observe the supported branch and exercise the serial path. Derive the row counts and documentation from that same contract. If there is a concrete reason to retain three UF2-only Firefox rows, the model and documentation still need names that express what those rows prove and a rationale for their distinct release value.

PR #114, opened 2026-08-14
usb_auto: release a port that never answers the handshake

jackspace:
I plugged an RNode into a Linux box that was running prnsd, and prnsd took `/dev/ttyUSB0` and held
it for twelve and a half minutes, until I physically unplugged the board. From the kernel log on
that host:

```
Aug 13 18:31:33  usb 1-1.4: cp210x converter now attached to ttyUSB0
Aug 13 18:31:34  cp210x ttyUSB0: usb_serial_generic_read_bulk_callback - urb stopped: -32
Aug 13 18:43:57  cp210x ttyUSB0: cp210x converter now disconnected from ttyUSB0
```

The middle line is one second after enumeration, before I could have reached the keyboard, and `-32`
is `EPIPE`. Everything else that wanted that port then got a hardware error that had nothing to do
with hardware:

```
[2026-08-13 18:33:17] [Error]  Could not detect device for RNodeInterface[RNode Rojo]
[2026-08-13 18:35:30] [Error]  The interface RNodeInterface[RNode Rojo] experienced an
                               unrecoverable error and is now offline.
```

and the probe I ran by hand to check the board was alive:

```
RNode did not respond. Is a valid firmware installed?
Got invalid response while detecting device
```

The board was fine. Neither tool mentioned contention, and prnsd didn't log anything at all, so it
took me a while to work out that three programs were fighting over one descriptor.

The scanner is doing what it's meant to do. `usb_auto` is a discovery interface, so of course it
opens candidates to find out what they are. The problem is that it never gives one back. A port
leaves the scanner exactly two ways today: the interface gets disabled, or the device disappears
from the scan. The port task itself only breaks on a write error or a read EOF, so a device that
accepts writes and never replies keeps `confirmed == false` indefinitely, and
`PortLiveness::Handshaking` never ages out of `LIVENESS_TIMEOUT` because that window only starts
counting after a handshake has landed.

So: give the handshake a deadline, and then don't immediately re-take the port once it fires.

I want to flag the second half, because the deadline on its own was my first attempt and it isn't
enough. With just the deadline and the existing 5 s `OPEN_FAILURE_BACKOFF` behind it, the scanner
releases the port and grabs it straight back, and the device is still claimed for most of every
cycle. In the test below that shows up as seven reopens where the finished version does three. So
`PortEvent::Closed` now carries whether the port ever confirmed, and `failed_opens` keeps that
unconfirmed count across reopens instead of clearing it on each retry. `reconcile`'s existing
`retain` already drops the entry when the port leaves the scan, so anything that re-enumerates gets
a clean slate, which is what I'd want after power cycling a board. Worth saying that the clean slate
depends on a scan actually observing the gap, so a replug that completes inside one
`FALLBACK_SCAN_INTERVAL` without the rescan signal firing would stay given up on. That's the
existing behaviour of `failed_opens` rather than anything new here, but it's now load bearing.

**Two numbers I'd rather you set than me.** `HANDSHAKE_DEADLINE` is 30 s and `MAX_UNCONFIRMED_OPENS`
is 3. I picked the deadline to sit comfortably beyond a board that resets and boots after its port
opens, since the probe's own comment says a newly opened board may still be booting, and abandoning
a real board mid-boot would be a worse bug than the one I'm fixing here. You have a much better
sense than I do of the slowest board that has to survive this, and of how patient discovery should
be. Both are single consts, so move them wherever you're comfortable.

One thing this deliberately does not do: it is not a device filter. Every USB serial device on the
host is still a candidate and still gets opened. This only bounds how long one is held. Whether
discovery should narrow what it opens in the first place seemed like your call rather than mine, so
I left it alone.

Tests are two new `#[tokio::test(start_paused = true)]` cases, one for the release and one for the
give-up. I checked they actually fail without the fix rather than assuming it: dropping the deadline
arm makes the first one time out and the second report a single open, and keeping the deadline while
dropping the give-up gate makes the second report seven.

Ran on Linux: `cargo test` and `cargo clippy -- -D warnings` for `prns-interfaces-tokio` with
`--features usb` (the module is behind that feature, so a bare `cargo test` builds none of it), plus
`validation/hygiene/fmt-docs.sh`, `validation/run.py verify` and `tools/prns verify`. The original
symptom was on Linux too. **I have not tested macOS**, where enumeration goes through `prns-ffi`,
and I don't have a Mac to try it on.

### DIFF
--- a/prns-interfaces/impls/tokio/src/usb_auto/mod.rs
+++ b/prns-interfaces/impls/tokio/src/usb_auto/mod.rs
@@ -40,6 +40,12 @@ const LIVENESS_TIMEOUT: Duration = Duration::from_secs(6);
 const RECENT_LINK_GRACE: Duration = Duration::from_secs(3);
 /// Backs off a busy or re-enumerating target so failures do not become a once-per-second error storm.
 const OPEN_FAILURE_BACKOFF: Duration = Duration::from_secs(5);
+/// Gives up a port that accepts writes and never answers, so a serial device that is not a board
+/// stops being held open. Long enough to cover a board that resets and boots after the port opens.
+const HANDSHAKE_DEADLINE: Duration = Duration::from_secs(30);
+/// Stops reopening a port that has never answered, until the entry re-enumerates. Without this the
+/// deadline alone still holds the device for `HANDSHAKE_DEADLINE` out of every backoff cycle.
+const MAX_UNCONFIRMED_OPENS: u32 = 3;
 
 struct Port {
     id: String,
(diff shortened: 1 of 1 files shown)


KenAKAFrosty:
A very welcome change, thanks!  Reminded me about a few items with the auto-USB I wanted to get back to, so I've added those to trunk in other recent commits as well

PR #118, opened 2026-08-16
host: keep -D warnings green where every control variant compiles out

jackspace:
Both Windows native lanes are red on trunk right now, and the break is one the Ubuntu lanes can't see. `NativeControl`'s only variant is `#[cfg(unix)]`, so on Windows the enum is uninhabited: `execute_control`'s parameters go unused and the host's `controls` sender is never read, and `-D warnings` promotes all three to errors.

Neither is a defect. The empty match is the totality proof over an uninhabited enum, and the control channel's plumbing is platform-neutral on purpose, so a future non-unix control slots in without re-plumbing. The fix scopes two allows to `not(unix)` with comments saying exactly that, rather than underscoring parameters the unix arm really does read.

Ran on Windows 11, `x86_64-pc-windows-msvc`, in `prns-host/impls/native`:

- reproduced the three CI errors on unmodified trunk first (`RUSTFLAGS=-D warnings cargo check`)
- after the fix: check and `clippy -- -D warnings` both green
- `RUSTFLAGS='-D warnings' cargo check` stays green on Linux (WSL), where the annotations are inert

### DIFF
--- a/prns-host/abi/c/src/supplied_pipe.rs
+++ b/prns-host/abi/c/src/supplied_pipe.rs
@@ -1,14 +1,19 @@
 use std::ffi::c_void;
 use std::ptr;
 use std::sync::{Arc, Mutex};
+// `Duration` and `issued_command` are only reached from the unix arms below; the non-unix arms
+// return `Unsupported` before either would be used.
+#[cfg_attr(not(unix), allow(unused_imports))]
 use std::time::Duration;
 
 #[cfg(unix)]
 use std::os::fd::{FromRawFd, OwnedFd, RawFd};
 
+#[cfg_attr(not(unix), allow(unused_imports))]
+use crate::issued_command;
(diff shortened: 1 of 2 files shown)


jackspace:
CI got further than my bench check did: with prns-host-native fixed, the lane reached prns-host-c and hit five more Windows-only warnings in supplied_pipe.rs, same mechanism (unix-only pipe backend, non-unix arms return Unsupported early). The trunk red had been masking that crate entirely. Second commit applies the same scoped allows there, verified with the lane's exact release-build command on Windows MSVC plus the Linux control.

KenAKAFrosty (approved):
Reviewed against current live trunk. The patch applies cleanly. Targeted Linux native check and clippy, plus the C ABI check, pass with warnings denied. Both hosted Windows MSVC lanes are green. The red deny aggregate is unrelated repository-baseline drift tracked separately.

PR #119, opened 2026-08-16
deny: follow the vendored SoftDevice bindings through the v6/v7 split

jackspace:
The deny lane is red on trunk: `nrf-softdevice-s140-v7 = 0.1.2-prns.1 is unlicensed`. The vendored SoftDevice bindings split into `-v6`/`-v7` crates at `0.1.2-prns.1`, but deny.toml's license exception and clarification still name `nrf-softdevice-s140@0.1.2`, so cargo-deny reports an unmatched exception and treats the new crate as unlicensed (both vendored crates declare `license-file` with no SPDX expression, which is exactly what the clarify existed for).

This points the exception and the clarification at the crates that exist now, one entry per binding. The pinned license-file hash is unchanged: both vendored LICENSE-NORDIC files are byte-identical to the original, so a future change to the license text still fails the audit.

Verified with cargo-deny 0.19.8 (the CI pin) using deps-audit.sh's exact flags:

- unfixed trunk, nrf52840 graph (`thumbv7em-none-eabihf`): reproduces the CI error, exit 4
- fixed, same graph: `licenses ok`
- fixed, engine graph (`x86_64-unknown-linux-gnu`): `licenses ok`, so nothing else moved


jackspace:
CI showed the license fix working and then hit the next two things the old red had been hiding: the unsafe dependency snapshot and the notices file were both stale (the snapshot predates the vendored split; the notices are one roster line behind). Second commit regenerates both with the audit script's own tools. One bench note in case anyone else regenerates these on Windows: a CRLF working copy fails the checksum-pinned license clarifies and relabels the Nordic license as BSD-3-Clause, so I verified from an LF checkout. Snapshot reports "matches reviewed baseline"; notices check is clean.

KenAKAFrosty:
Thanks for the continued help with all the last-mile edges!

PR #124, opened 2026-08-19
Keep the limits rows inside the panel

jackspace:
The Limits screen writes counts as raw digits while every other number on the face goes through
`fmt_count`, and nothing checks that a row fits. On `GrowableHeap`, `packet_hashes` is the one
`Fixed` capacity and the largest number anywhere in the UI, so the row renders as `PktHash 500000`.

That is fourteen FONT_5X8 characters, drawn from `x = 2` on a 64 pixel panel, which draws twelve:

    P k t H a s h _ 5 0 0 0 | 0 0
    1 2 3 4 5 6 7 8 9 . . 12| 13 14

The thirteenth character is the fourth zero, and that is exactly where it is sliced. I set the
Android build up on a cheap video projector running Android 11 on 1 GB of RAM, pointed it at a
wall, and read the screen.

The root cause is that the row is assembled into an `HString<16>`, four characters wider than the
panel can draw, and nothing reconciles the two.

**The boards never showed it.** Their layouts declare capacities like 36, 64 and 1, and `fmt_count`
leaves anything under a thousand alone, so no embedded reading changes here at all. Seeing this needs
the Limits screen on a heap platform: desktop, Android or iOS.

Two things move:

- `LimitValue::Count` renders through `fmt_count`, so the row becomes `PktHash 500K` at twelve
  characters and fits exactly.
- The row string now comes from one shared `limits_row_text`, so the test asserts the same bytes the
  screen draws rather than its own idea of them.

The guard sweeps `packet_hashes` across every magnitude a `u32` can hold rather than pinning the
500,000 that broke, so it cannot quietly stop covering `GrowableHeap` if that constant moves. It
also covers an all-dynamic layout and a board-sized one, and a second test pins that a small
capacity is still rendered exactly, because compacting to fit must not cost precision where there
is room for it.

Proven red before it was trusted green. Reverting the one `fmt_count` call fails it with:

    row "PktHash" renders 14 characters as "PktHash 500000", panel draws 12

Remaining headroom, named rather than left to be rediscovered: `Receipts` is the longest label at
eight characters, so it fits a three character value and no more. Nothing declares a four figure
receipt table today. If something does, this is the test that will say so.

Run: 139 tests in `personal-hopspot-core` and `cargo clippy --all-targets -- -D warnings` clean, both
on Ubuntu under WSL, which is what CI runs; `VALIDATION_REGISTRY_OK`, `TOOLING_REGISTRY_OK` and
`FMT_DOC_CHECK_GATE_OK` from `validation/hygiene/fmt-docs.sh` on Windows; and the `heltec-v4-r8`
firmware builds, also on Windows, since this code compiles into it.

### DIFF
--- a/personal-hopspot/core/src/screen/render/layout.rs
+++ b/personal-hopspot/core/src/screen/render/layout.rs
@@ -53,6 +53,8 @@ pub(super) const MENU_TEXT_X: i32 = 12;
 pub(in crate::screen) const MENU_REASON_X: i32 = 2;
 pub(super) const MENU_DETAIL_STEP: i32 = 7;
 pub(in crate::screen) const FONT_5X8_CHAR_W: i32 = 5;
+
+pub(in crate::screen) const LIMITS_TEXT_X: i32 = 2;
 pub(in crate::screen) const FONT_4X6_CHAR_W: i32 = 4;
 pub(super) const GLOBAL_MENU_ITEM_STEP: i32 = 11;
 pub(in crate::screen) const SPLASH_TEXT_X: i32 = 2;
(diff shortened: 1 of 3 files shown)


KenAKAFrosty:
Yep good catch here. Made a few modifications but kept the spirit of this. Thanks!

PR #126, opened 2026-08-20
Let Android Hopspot export a live RNS join config

tonycowan:
## Summary
- Persist a per-device RPC key on Android Hopspot and serve it on the shared-instance RPC port.
- Copy a Sideband-style Reticulum template from the Local card **RNS Config** action so other apps can join without a hardcoded key.
- Keep the menu item behind `android-rns-config` so iOS, desktop, and embedded stay Power/Back only.

## Test plan
- [ ] Hopspot core tests with and without `android-rns-config`
- [ ] Android: Local card → long-press **RNS Config** copies the commented template with the live `rpc_key`
- [ ] Stock Sideband: paste into Utilities → Advanced Reticulum settings and confirm Full RNS Status joins Hopspot
- [ ] Non-Android Hopspot Local menus do not show **RNS Config**

### DIFF
--- a/personal-hopspot/core/Cargo.toml
+++ b/personal-hopspot/core/Cargo.toml
@@ -14,6 +14,7 @@ build = "build.rs"
 default = []
 host = ["personal-rns/tokio-host"]
 source-archive = ["personal-rns/large-static-responses"]
+android-rns-config = []
 
 [dependencies]
 personal-rns = { path = "../../personal-rns", default-features = false, features = ["flash"] }
--- a/personal-hopspot/core/src/mobile.rs
+++ b/personal-hopspot/core/src/mobile.rs
@@ -46,6 +46,7 @@ impl InvalidMobileInputCode {
 pub enum MobileActionCode {
     None = 0,
     Announce = 1,
+    CopySidebandJoinConfig = 2,
 }
 
 impl MobileActionCode {
@@ -53,6 +54,7 @@ impl MobileActionCode {
     pub const fn encode(action: UiAction) -> Self {
         match action {
             UiAction::Announce => Self::Announce,
+            UiAction::CopySidebandJoinConfig => Self::CopySidebandJoinConfig,
             UiAction::None
             | UiAction::OledOff
             | UiAction::ToggleOledAutoOff
@@ -226,6 +228,10 @@ mod tests {
         assert_eq!(
             MobileActionCode::encode(UiAction::Announce),
             MobileActionCode::Announce
+        );
+        assert_eq!(
+            MobileActionCode::encode(UiAction::CopySidebandJoinConfig),
+            MobileActionCode::CopySidebandJoinConfig
         );
         assert_eq!(
             MobileActionCode::encode(UiAction::Sleep),
--- a/personal-hopspot/core/src/screen/render/menus/mod.rs
+++ b/personal-hopspot/core/src/screen/render/menus/mod.rs
@@ -43,6 +43,26 @@ pub(in crate::screen) const fn station_uplink_action_label(kind: CardKind) -> Op
     }
 }
 
+pub(in crate::screen) fn interface_menu_item_label(card: &Card, index: usize) -> &'static str {
+    let items = interface_menu_items(card.kind);
+    let item = items.get(index).copied().unwrap_or(items[items.len() - 1]);
+    if index == POWER_MENU_ITEM {
+        if card.connection == ConnectionState::Disabled {
+            "Turn On"
+        } else {
+            "Turn Off"
+        }
+    } else if matches!(
+        card.kind,
+        CardKind::WifiStation | CardKind::WifiStationDisabled
+    ) && index == STATION_UPLINK_MENU_ITEM
+    {
+        station_uplink_action_label(card.kind).unwrap_or(item)
+    } else {
+        item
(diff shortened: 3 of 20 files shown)


KenAKAFrosty:
Hey thanks Tony! Made some refinements and tweaks and all's in a good place now for what you're looking to use it for, as much as it can be on our end at least!

jackspace:
Ran the test plan on hardware tonight. First phone: Samsung S24 Ultra, Android 16 (API 36), Hopspot built from `main` at `376bf1a1`, stock Sideband alongside it.

- Local card, long-press **RNS Config**: copied the commented template with the live `rpc_key`, as described.
- Sideband, Utilities, Advanced Reticulum settings, paste, save, full restart. Full RNS Status then listed the Hopspot's interfaces (AutoWifi, BluetoothAuto, WifiDirect, WifiAware, and the USB Auto `Interface[d0d0d0d0]`), `Shared Instance[09699a59]` Up with traffic both ways, and the Hopspot's transport identity as the running transport instance. Sideband's own log shows `Loading modified RNS config template`, and the Hopspot's Local card showed traffic both ways on its shared instance.
- Did it twice on the same phone: first with Sideband 1.7.0 (ships RNS 1.0.0, so pickle RPC), then with 2.0.1 installed over it (ships RNS 1.4.2, msgpack). Same status page both times, and the pasted config survived the upgrade. One wrinkle from logcat: on 1.7.0 the in-app full restart only relaunched the UI process, so Sideband's background service kept its pre-paste config until the 2.0.1 install killed it, and its RPC calls to the Hopspot got `digest sent was rejected` in the meantime, which is the key check doing its job. On 2.0.1 both processes loaded the pasted template.

Then the same on a Pixel XL (LineageOS, Android 15, API 35) with Sideband 2.0.1 and a freshly installed Hopspot: same status page, and a useful before/after in its log, `digest sent was rejected` on the RPC before the paste, `Loading modified RNS config template` from both the UI and the service after it.

Two things I didn't see: which dialect each round actually used on the wire (the Hopspot logged nothing at its default level), and I didn't push LXMF traffic through the joined instance. Screenshots and the logcat excerpts are on hand if useful. No retries needed; thanks for this one, @tonycowan.

tonycowan:
Cool! Thanks for your testing @jackspace. If you're doing more android work while we wait for that nice UI that's on its way, you may be interested in this pretty version of hopspot for android. Seems to work for me: https://github.com/tonycowan/Prns/tree/feat-android-dioxus-demo/personal-hopspot/mobile/dioxus-android
Apk available if you don't want to build it.

jackspace:
Thanks @tonycowan, this is nice work. I pulled your branch this afternoon, built it here (dx 0.7.5, Windows), and ran it on a Pixel XL over the live service. The bridge works: real node state in the touch UI within seconds, and the service kept running underneath everything I did to the UI, exactly the design you described.

I think I found the root cause of the freeze f684ac31 was chasing, and it is not the clipboard. `ui.rs` and `model.rs` time the toasts with `std::time::Instant`, which panics on wasm32-unknown-unknown; the panic string is baked into the committed wasm, and the toast-bearing presses (Announce, the toggles, the config copy) all hit it. On my bench the first tap did its work (the announce went out, I watched TX jump) and then trapped with `Uncaught RuntimeError: unreachable` in logcat, no toast; the next tap, on More, which never even makes a toast, trapped again and the UI froze for good, uptime stuck on screen while the service kept running underneath. Swapping in `web_time::Instant` (or `js_sys::Date::now()`) should fix it wholesale.

Two more things you will hit before CI goes green, both quick: `python validation/run.py verify` fails because `dioxus-android`'s Cargo.toml and lock are not in `validation/manifest.toml`, and the `oled` flavor cannot compile because `DioxusHostView.kt` lives in `src/main` while androidx.webkit is `dioxusImplementation`-only (I got `Unresolved reference: webkit` on `assembleOledDebug`; the flavor's release-runtime tsv will want the webkit coordinates too). One design note since you built on an older base: main now has `engine::sideband_join_config()` from this PR with a tested canonical template, so your hand-rolled one in `ui_live.rs` can become a one-line call when you rebase; main also sets the sensitive-clipboard flag on the MainActivity copy path, worth mirroring in DioxusHostView's.

I have a longer list of smaller notes (the snapshot dedup never fires because uptime_ms churns every poll, the rpc_key rides every snapshot though the UI never reads it, WebView remote debugging is on unconditionally, back gesture exits the app from detail screens). Happy to drop them on your branch if you want them.

jackspace:
One more, because this deserves eyes and not just bug reports: here it is on a 2016 Pixel XL against the live node, with my S24 sitting next to it.

<hr>

01-home

<hr>

04-lan-detail

<hr>

05b-ble-detail

<hr>

That BLE peer is the S24, phone to phone. The whole thing is clean and quick even on nine-year-old hardware, and it reads like a product, not a demo. Really nice work @tonycowan!

tonycowan:
I'm always impressed by the thoroughness of your tests and feedback @jackspace ! Thank you for your time and efforts. I built this just for us to play with while Ken puts together the management UI and there are many things (mostly surfacing extra data) that I would like to add to it if it had a future, but everything it does will be a subset of what the upcoming management UI will do, so I want to focus my efforts on other things for now.

PR #128, opened 2026-08-21
interfaces: let a family account for the frames its RX path discards

jackspace:
You asked for this on #103:

> My preference is to extract the typed reassembly improvement, then supersede the transport and
> reporting portions with the unified Remote Management that will begin extremely soon.

So here it is on its own, with nothing else attached. You have since closed #103 pointing at #131,
so read this as that extraction rather than as a piece of the old thread. It carries none of the USB
transport and none of the vitals message.

### What it is

Byte counters can't tell "nothing arrived" from "something arrived and was thrown away". A frame
dropped before reassembly has already moved `rx_bytes`, so a link discarding everything looks exactly
like a healthy idle one from outside the node. That's the failure I sat and watched for hours this
month: a wedged radio module passing zero frames while every counter a host could poll read fine.

`FrameAccounting` is four counters (`frames_in`, `malformed`, `undecodable`, `delivered`) hanging off
`InterfaceStatus` as a defaulted method, so no existing family has to change. The embassy status backs
them with atomics behind an opt-in declaration.

### The part I care most about

`frame_accounting()` returns `Option`, and `None` means the family does not count. It does not mean
zero. Those are different answers and collapsing them is how a reporting layer starts lying: an
unaccounted interface would read as a perfectly idle one. A driver that wants accounting calls
`account_frames()` once at bring-up, before the first frame, and the counters publish from then on.
Families that never declare stay `None` forever, and to be plain about it: nothing declares yet.
This PR is the typed shape only, so every family reports `None` until a driver opts in, which is the
correct reading of the current tree.

There's a test for exactly that, `frame_accounting_stays_unpublished_until_declared`: it counts frames
before declaring and asserts the reading is still `None`. Without it, "unaccounted" quietly becomes
"zero" the first time someone adds a counter call in the wrong order.

One difference from #103 worth naming so you don't have to diff for it: that branch carried the four
counters plus `last_frame_in_at_ms`. This ships the counters only. The timestamp is the part that
drags in sampling semantics, which is exactly what you said should be shaped by the unified model,
so it stays behind.

I also tightened one doc comment since I opened this. `delivered` used to say the frame was "handed
upward to the engine". The counter is incremented at the handoff, so it can't speak for what the
engine then does with the frame, and on a link where packets arrive and produce nothing that
distinction is the whole question. It now says it counts the handoff and not the engine's verdict.
A struct whose entire job is separating arrival from discard shouldn't overstate one of its own four
numbers.

### What's deliberately not here

No transport, no wire format, no capability negotiation, no USB. Those are the parts you said should
be superseded by the unified interface report, and I agree they shouldn't be settled in a USB-only
message. This is only the typed shape the reporting model would sit on top of.

`InterfaceSnapshot` is untouched. I left it alone on purpose since it's yours to decide whether the
accounting belongs in the snapshot or beside it, and that decision reads differently once Remote
Management exists.

### On the six red checks that were sitting here

They weren't from this change. This branch was 124 commits behind, and the red was two separate
pieces of base breakage.

The platform jobs were the 1.98 `chunks_exact` lint in the excluded workspaces: `windows-desktop`
failed on `error: using chunks_exact with a constant chunk size` in `prns-ffi`, and
`macos-ble`, `android-service` and `feature-configs` went with it. You've since fixed that on
trunk, and merging trunk in cleared all four here.

`deny` and the `Release critical` roll-up are the other piece, and they are still red: `unsafe
dependency snapshot drifted; review and run validation/security/unsafe-audit.py --write`. That is
not something this branch can fix, and it isn't inherited-and-stale either. Trunk's own CI run on
`55857a75`, the commit I merged, fails on exactly those two jobs right now. Flagging it in case it
is news.

### Verification

Three files changed, +108 / -3. Re-run today against the merged result, not carried over from when I
opened this.

- `bash validation/hygiene/fmt-docs.sh` -> `FMT_DOC_CHECK_GATE_OK`
- `python validation/run.py verify` -> `VALIDATION_REGISTRY_OK`
- `./tools/prns verify` -> `TOOLING_REGISTRY_OK`
- `cargo clippy -p prns-core --all-targets -- -D warnings` -> clean
- in `prns-runtime/impls/embassy`: `cargo clippy --all-targets --locked -- -D warnings` -> clean, and
  `cargo test --locked` -> **80 passed, 0 failed**

I ran `frame_accounting_stays_unpublished_until_declared` by name as well as in the suite, and got
`1 passed; 79 filtered out`, because the total alone wouldn't prove it executed.

That matters more than it sounds. `prns-runtime/impls/embassy` is in the root workspace's `exclude`
list, so `cargo test --workspace` from the root never runs that test at all: it passes without ever
having seen it. The assertion only executes if you run the tests from inside that workspace.

Same caveat for the other implementors. The concrete `InterfaceStatus` impls live in the
`prns-interfaces` and `prns-runtime` impl workspaces, all excluded from the root, and of those I
linted and tested only `prns-runtime/impls/embassy`, the one this change touches. The method is
defaulted, so the others compile unchanged by construction, but I'm stating what I ran rather than
implying the rest were checked.

Host is Ubuntu 24.04 under WSL, rustc 1.98.0. I did not run macOS or iOS.


KenAKAFrosty:
Heck yeah, this is a great foundation! Thanks for this

KenAKAFrosty:
Also fun little bonus! @jackspace got this wired into the OTEL now so there's already a target to consume it.  be85ed0199a59e2e50a88bb9b605e1baf7968461 

Took that chance to also add protocol violations and expand that, so there's some extra data in there now too! Thanks again for this one!

PR #137, opened 2026-08-22
benchmarks: pass the route evidence id the route tables now require

jackspace:
`benchmarks` doesn't build on `main` right now. `RouteTable::push` took a `RouteEvidenceId` when the
route evidence ids landed (`prns-core/src/routing/routes/core.rs:146`), and two benches still call
the old two-argument form: `benches/route_interface_index.rs:42` and `benches/route_index.rs:47-48`.
Nothing in CI builds that workspace (it detaches itself with an empty `[workspace]`, by design), so it
went red quietly.

The fix hands each row a distinct id the same way the core's own table tests do
(`RouteEvidenceId::new(n + 1)`, the helper in `prns-core/src/routing/routes/impls/fixed_array.rs`),
in both files. Nothing else
changes; the measured paths (`index_of`, `repoint_receiving_interface`) don't read the id.

What I ran, WSL Ubuntu 24.04 on Windows 11, rustc 1.98.0, inside `benchmarks/`:

- Control on unmodified `main` `b519d014`: `cargo check --all-targets --locked` exits 101 with
  `E0061: this method takes 3 arguments but 2 arguments were supplied` at
  `route_interface_index.rs:42`.
- After: the same command exits 0.
- `cargo bench --locked --bench route_index --bench route_interface_index -- --test` (criterion's
  test mode, each bench body once): exit 0, all 36 bench points report `Success`
  (`route_lookup_hit`/`miss` from 16 to 1024 rows, `route_repoint_no_match` and
  `route_repoint_sparse` from 1,000 to 1,000,000 rows). So the benches run, not just compile.
- Repo gates from Git Bash: `FMT_DOC_CHECK_GATE_OK`, `VALIDATION_REGISTRY_OK`, `TOOLING_REGISTRY_OK`.

One thing I didn't touch: `cargo clippy --all-targets --locked -- -D warnings` in the same workspace is
still red for three pre-existing sites in `src/bin/raw_transport_driver.rs` (`ptr_arg`,
`redundant_pattern_matching`, `too_many_arguments`). They're outside this fix, so I left them alone
and mention them only so nobody reads this as "benchmarks is lint-clean now".

No macOS or Linux-native run; WSL and Windows only.


KenAKAFrosty:
Maintainer revalidation after retargeting this PR to trunk and updating the branch at head 9dc6bfffacf4d473edb9a0e46ec71fa39336ae53:

Host: macOS (Apple Silicon), repository Rust toolchain.

- cargo check --all-targets --locked in benchmarks: passed.
- cargo bench --locked --bench route_index --bench route_interface_index -- --test: passed.
- All 36 Criterion test-mode points reported Success across route lookup hit/miss and route repoint no-match/sparse cases.

This verifies the detached benchmarks workspace itself against the exact updated PR head while the protected repository checks run.

KenAKAFrosty:
Thank you, Jack—this was a sharp, focused catch. The detached benchmarks workspace had quietly drifted from the RouteTable API, and this restores both compilation and the intended benchmark paths without changing what they measure. Merged into trunk with your authorship preserved.

PR #141, opened 2026-08-23
hopspot: show the battery percentage instead of quantized bars

jackspace:
The gauge already knows the exact number. `BatteryGauge` folds VBAT into a smoothed 0 to 100 percent, and then the title bar rounds it to the nearest quarter and draws four segment bars. At the bottom of the discharge, that quantization hides exactly the thing you look at the panel to learn.

What put this on my bench: I ran a V4-R8 on a 3000 mAh pack for 23 hours, walked up to it, and the panel showed an empty outline. The board was fine, and some real percent was sitting in the gauge, but the display could not tell me whether I had 2% or 12%, which is the difference between "grab the charger now" and "it can wait until tonight." I plugged it in blind.

The change renders the percent itself, right-aligned in the same 20 px the glyph occupied ("100%" in FONT_5X8 is exactly that wide). It reads the `PowerSnapshot` the renderer already gets: while external power is present and the cell is below full, a compact plug cue sits at the left of the zone, steady when power is merely present and blinking on the existing 600 ms cadence while actually charging; at 100% the cue drops and the full-width number reads as done. Unknown renders `--%`. No layout, state, or gauge changes, so every surface the shared renderer drives (OLED, eink, desktop, the mobile panels) picks it up together.

Tests were rewritten to match: five of the six compare the whole zone against an image produced by the same text draw the implementation makes, so they pin alignment and content without hand-encoded font bitmaps; the sixth probes the plug cue pixels. They cover level, right-alignment as digits shrink, unknown, both charging blink phases, and the 100% no-cue case. At this commit I widened the zone by one pixel and watched five of the six fail before trusting them.

Run here, all at this commit: `cargo test -p personal-hopspot-core --locked` (147 passed) and `cargo clippy -p personal-hopspot-core --all-targets --locked -- -D warnings` on Ubuntu 24.04 at rustc 1.96.0, the toolchain your CI pins; `bash validation/hygiene/fmt-docs.sh` (FMT_DOC_CHECK_GATE_OK) in Git Bash on Windows; the shared-renderer consumers built clean with zero warnings: `cargo heltec-v4-r8` (esp toolchain, Windows), the t-echo firmware with `board-t-echo` on both `softdevice-s140-v6` and `softdevice-s140-v7` (your CI step, WSL), and `cargo clippy --all-targets -- -D warnings` on the desktop face (WSL). On hardware: the Android Hopspot built from this branch runs on an S24 Ultra and a Pixel XL; on the S24 the title bar read 53% against Android's own 53. Not run: on-device rendering on the T-Echo eink or the V4-R8 OLED itself; the mock-display tests cover the renderer, and I will put it on V4-R8 hardware here as soon as you want a photo.

### DIFF
--- a/personal-hopspot/core/src/screen/render/glyphs.rs
+++ b/personal-hopspot/core/src/screen/render/glyphs.rs
@@ -12,7 +12,7 @@ use super::layout::*;
 use super::primitives::{draw_pattern_colored, fill, line, line_colored, stroke};
 
 /// How wide the title-bar battery zone is: "100%" in FONT_5X8 exactly.
-const BATTERY_ZONE_W: i32 = 20;
+const BATTERY_ZONE_W: i32 = 4 * FONT_5X8_CHAR_W;
 
 /// The battery readout, drawn in the background color (it sits on the inverted title bar): the
 /// percentage as text, right-aligned in a 20px zone so the digits do not wander as they shrink,
@@ -35,7 +35,7 @@ pub(in crate::screen) fn draw_battery<D: DrawTarget<Color = BinaryColor>>(
             let _ = text.push_str("--%");
         }
     }
-    let text_x = x + BATTERY_ZONE_W - text.len() as i32 * 5;
+    let text_x = x + BATTERY_ZONE_W - text.len() as i32 * FONT_5X8_CHAR_W;
     let small = MonoTextStyle::new(&FONT_5X8, BinaryColor::Off);
     let _ = Text::with_baseline(&text, Point::new(text_x, y), small, Baseline::Top).draw(display);
     // External power keeps a compact plug cue at the left of the zone while the cell is below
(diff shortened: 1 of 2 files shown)


jackspace:
Photos of the two Android cases from the body, since they are easier to check than my description of them.

S24 Ultra, charging, at 53%. The title bar reads 53% and Android's own indicator reads 53 in the same frame, so you can see the two agree without taking my word for it. The plug cue is the small block just left of the digits.
<hr>
pr141-s24-53pct
<hr>
Pixel XL, on USB at 100%, same build. The cue is gone and the number runs the full width, which is the 100% no-cue case the tests pin.
<hr>
pr141-home
<hr>

Still nothing from me on a V4-R8 OLED or a T-Echo eink. Happy to put it on the board here whenever you want that photo.

KenAKAFrosty:
A nice little touch!  It got a bit blended in that top header without the battery outline, so I added that back. I sacrificed the '%' sign since the context helps you understand anyway (in fact, your Android screenshots show it doing the same thing haha!) Thanks for this

jackspace:
Even better!  That percent sign ( % ) was kind of a waste of screen real estate.

Thanks for being so responsive @KenAKAFrosty !

PR #165, opened 2026-08-26
Surface nested Auto fleet peers in prnsd status

tonycowan:
## Summary
- Preserve Auto fleet members (Bluetooth, Wi-Fi, Weave) through inventory fold instead of dropping them when aggregating supervisors
- Name peers at fleet admission and carry optional link-up RSSI for Bluetooth peers
- Encode nested `fleet_peers` on interface status and render them under each Auto supervisor in `prnsd status` human and JSON output

## Test plan
- [x] Fold inventory unit tests (members retained with names/bytes/RSSI)
- [x] MessagePack encode/decode round-trip for `fleet_peers`
- [x] rnstatus human/JSON render tests
- [ ] `prnsd status` with a connected Bluetooth Auto peer shows nested peer name, Up/Down, RSSI, and traffic
- [ ] `prnsd status -j` includes `fleet_peers` under each Auto supervisor

### DIFF
--- a/prns-core/src/interfaces/bluetooth_auto/policy.rs
+++ b/prns-core/src/interfaces/bluetooth_auto/policy.rs
@@ -62,6 +62,7 @@ pub enum PolicyAction {
         slot: usize,
         address: BleAddress,
         lane: L2capPlan,
+        peer_rssi: Option<i8>,
     },
     Evict {
         identity: BleIdentity,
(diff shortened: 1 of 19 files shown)


tonycowan:
Closing in favor of #171, which rebases this work onto trunk together with BLE discovery groups and the related status plumbing (group_id, software_version, fleet-member fold fix).

No need to merge both — #171 already contains the nested Auto peer status commits.

PR #166, opened 2026-08-26
Abstract display semantics to avoid duplication across boards

evelant:
Refactors and abstracts display handling to better support eink and more devices / displays without duplicating code for each one. Also lays groundwork for supporting display formats other than 64x128 portrait. 

This is specifically because I've got another pr to support the [Heltec E290](https://heltec.org/project/vision-master-e290/) which is ESP32-S3 based but has a larger eink display. The current esp32-s3 implementation was oled only. I didn't want to duplicate the eink, rotation, or other code that was already added for other boards.

Sorry this PR is so big. I couldn't think of a good way to do this in smaller pieces without adding temporary compatibility shims which seemed a worse option. If this is a useful direction I will need some help testing because I don't own any of the other boards at the moment so I can't be completely sure I didn't break anything on them despite all tests passing. 

Heltec E290 PR: #167


evelant:
Rebased on latest trunk and fixed the CI issues

evelant:
@KenAKAFrosty it looks like in some recent commits you did a lot of the same things that I did in this PR. Should I close this and redo #167 on latest trunk?

KenAKAFrosty:
VERY welcome refactor!! I tuned it a bit so the shape is different from where you started, but still overall this is a very welcome change and a bit overdue. Thanks for kicking this off!!

PR #167, opened 2026-08-26
Heltec e290 support

evelant:
Adds support for [Heltec Vision Master E290](https://heltec.org/project/vision-master-e290/). Stacks on top of #166 

Tested with two boards. Display, wifi, espnow, lora, ble, and usb all working well.

PR #168, opened 2026-08-27
fix(hopspot): flash through the board's own partition table

jackspace:
The four `-flash` aliases in `personal-hopspot/embedded/esp32/.cargo/config.toml` run espflash with no `--partition-table`, so espflash uses its built-in default. That is not the layout the firmware is built against.

I dumped both tables off the same ELF rather than reasoning about it, with `espflash save-image --chip esp32s3 --flash-size 16mb --merge` once with the flag and once without, then decoded the table at `0x8000` from each image:

```
built-in default                  partitions-hopspot-16mb.csv
nvs      0x009000 +0x006000       nvs        0x009000 +0x003000
                                  ble_id     0x00c000 +0x001000
                                  hopcfg     0x00d000 +0x001000
                                  node_id    0x00e000 +0x001000
phy_init 0x00f000 +0x001000       phy_init   0x00f000 +0x001000
factory  0x010000 +0xfa0000       factory    0x010000 +0xe6e000
                                  radio_cfg  0x0e7e000 +0x002000
                                  prns_state 0x0e80000 +0x180000
```

The default `nvs` covers `ble_id`, `hopcfg` and `node_id`. The default `factory` covers `radio_cfg` and `prns_state`. On a fresh board that costs nothing. On a board that already carries an identity, flashing through the documented alias takes `node_id` with it, and that one is not recoverable.

The fix is to give each alias its board's own table. The tables are the same files the flash manifest catalog already names per board, so the aliases now agree with the source of truth that `personal-hopspot/flasher`, the web flasher and `release.firmware.build` were already using. This was the one path that bypassed it. That is also why it is not a single `espflash.toml`: `heltec-v4-r8` and `t-beam-supreme` share the `xtensa-esp32s3-none-elf` runner but need different tables, so one file at that level would be wrong for one of them.

I added a check as well, because fixing the four aliases would have left the door open behind me. Two open pull requests add new `-flash` aliases and both are missing the flag the same way: #167 adds `heltec-e290-flash` and #95 adds `heltec-v3-flash`. Whichever of us lands second, the aliases that arrive would be unprotected.

The check reads the config, matches every `-flash` alias to a board on the catalog's own `package` field, and asserts the alias passes exactly the `partition_table` that board declares. Nothing is written down twice: add an ESP board to the catalog and the count assertion demands an alias for it, change a board's table and the alias has to follow. It sits next to `esp_partition_tables_match_the_firmware_layout_contract`, which already holds the CSVs to the firmware layout. This holds the aliases to the CSVs.

I checked that it fails on all three ways this goes wrong rather than assuming it would: a missing flag, a table belonging to a different board, and a catalog board with no alias at all. Each one fails with the offending alias named.

Two details on the fix itself. Paths resolve from `personal-hopspot/embedded/esp32`, which is where the aliases live and, since cargo discovers them by walking up from the working directory, the only place they can be run from. And if a path ever stops resolving, espflash exits non-zero and writes no image, so this fails loudly instead of quietly falling back to the default.

Gates on the branch: `FMT_DOC_CHECK_GATE_OK`, `VALIDATION_REGISTRY_OK`, `TOOLING_REGISTRY_OK`, `cargo clippy -- -D warnings` clean, and `cargo test` at 2064 passed / 0 failed across 26 suites against 2063 on the base commit, so this adds exactly the one test and drops nothing. I verified the aliases actually carry the flag by pointing `CARGO_TARGET_XTENSA_ESP32S3_NONE_ELF_RUNNER` at a script that echoes its arguments, so no board was flashed to check it:

```
RUNNER-ARGS: ...\hopspot-heltec-v4-r8 --partition-table partitions-hopspot-16mb.csv
```

This is the loose end from the last part of my comment on #46.

### DIFF
--- a/personal-hopspot/embedded/esp32/.cargo/config.toml
+++ b/personal-hopspot/embedded/esp32/.cargo/config.toml
@@ -3,6 +3,12 @@
 #   cargo heltec-v4-r8     / -flash   the Heltec V4-R8 (S3R8 / Octal PSRAM)
 #   cargo tbeam-supreme    / -flash   the LilyGO T-Beam S3 Supreme
 #   cargo c6               / -flash   the headless XIAO ESP32-C6
+#
+# Each -flash alias passes the board's partition table explicitly. Without it espflash
+# falls back to its built-in default, whose `nvs` spans ble_id/hopcfg/node_id and whose
+# `factory` spans radio_cfg/prns_state, so flashing a provisioned board through these
+# aliases would take its identity with it. The tables are the same files the flash
+# manifest catalog names for each board, and paths resolve from this directory.
 
 [target.xtensa-esp32s3-none-elf]
 runner = "espflash flash --monitor --after watchdog-reset"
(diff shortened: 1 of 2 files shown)


jackspace:
Closing this myself, because the central claim in it is wrong and I would rather say so than leave it standing.

I wrote that flashing through the `-flash` aliases takes `node_id` with it. I then put it on hardware and it does not. Same ELF flashed twice on a V4-R8 carrying a real identity, only the partition table differing:

```
--partition-table given   App/part 1,832,800/15,130,624   boots   identity unchanged
no flag                   App/part 1,832,800/16,384,000   boots   identity unchanged at +65s and +10min
```

Board boots normally either way, and all three identity partitions were byte-identical after ten minutes with the app confirmed running, then again after restoring the correct table. espflash writes only the images it is handed, so it does not erase a region merely because the new table declares it as `nvs`.

What is left is narrower and I have not demonstrated harm from it: the aliases write a partition table that does not match the one the firmware is built against, so the identity partitions stop being resolvable by name. That is a real mismatch and the fix is a small one, but it is not the data-loss hazard I opened this with, and the fix has a cost of its own - it makes the aliases sensitive to the directory they are run from.

I will raise the mismatch on discussion #46, where I first mentioned it, so it can be judged on what it actually is. Sorry for the noise.

PR #173, opened 2026-08-28
Embassy Wi-Fi Auto: unicast keepalive for known peers

tonycowan:
## Summary
- Refresh known Wi-Fi Auto peers with unicast peering tokens on embassy, every three beacon cycles (~4.8s), matching the tokio host cadence.
- Keeps LAN links alive on APs that block IPv6 link-local multicast peering beacons while mDNS discovery still works.

## Test plan
- [ ] Build an embassy Wi-Fi Auto board image (Heltec V4-R8 or T-Beam Supreme)
- [ ] Join a LAN with a Mac/Android tokio peer on the same discovery group
- [ ] Confirm peers stay connected past the 22s peering timeout without multicast beacons
- [ ] Confirm non-R8 S3 boards still build (`cargo heltec-v4`)

### DIFF
--- a/prns-interfaces/impls/embassy/src/wifi_auto/mod.rs
+++ b/prns-interfaces/impls/embassy/src/wifi_auto/mod.rs
@@ -958,6 +958,14 @@ impl<'a, const MEMBERS: usize> AutoWifi<'a, MEMBERS> {
                 self.status,
             )
             .await;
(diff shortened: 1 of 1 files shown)


tonycowan:
Closing in favor of #179, which already includes this unicast keepalive plus the IPv4 mDNS / discovery-path fixes for the same AP failure mode.

PR #175, opened 2026-08-28
Fix ESP32 Hopspot storage capacity for RemoteControl

tonycowan:
## Summary
- Raise ESP32-S3 held-identity capacity from 2 to 3 so RemoteControl (controller + target) and the Hopspot node identity can coexist during recipe assembly.
- Raise upstream-app destination capacity from 2 to 3 for the RC target, lxmf/delivery, and node/page destinations.
- Reserve one extra request-handler row on embedded Hopspot builds for the RC endpoint in addition to node-page routes.
- Add host tests that reproduce the old limits and verify the Hopspot + RemoteControl recipe assembles cleanly at the new capacities.

## Root cause
Merging RemoteControl into the Hopspot ESP32 firmware left core 1 panicking inside `init_static_with_persistence` (`HoldIdentity(StoreFull)` / `Register(RegistryFull)` / request-handler table full). Core 0 kept running (WiFi associated, `network.ready`), but core 1 never reached the liveness task, producing the ~15s `watchdog: core1 heartbeat missing` reboot loop on Heltec V4 R8.

## Test plan
- [x] `cargo test hopspot_style_recipe` in `prns-runtime/core`
- [x] `cargo check -p hopspot-heltec-v4-r8 --target xtensa-esp32s3-none-elf` (ESP32-S3 compile-time asserts)
- [x] Hardware: Heltec V4 R8 flashed from integration branch — 90s stable, manifold events, zero watchdog warnings

Made with [Cursor](https://cursor.com)

PR #177, opened 2026-08-28
Size embedded storage for RemoteControl

evelant:
RemoteControl consumes two held identities, one upstream application destination, and one request handler. The fixed embedded Hopspot profiles did not reserve all of that capacity, causing E290 startup to panic with HoldIdentity(StoreFull).

This change adds explicit RemoteControl capacity requirements, applies them to the ESP32-S3, ESP32-C6, and nRF52840 storage profiles, and adds an exact-capacity regression test.

Validated with the affected test suites, cross-target release builds, and an E290 hardware boot smoke.


tonycowan:
Thanks @evelant ! Looks like we both bumped into this. Might be some overlap here with #175.

PR #178, opened 2026-08-28
Add Montreal LoRa profile with per-board TX ceilings

tonycowan:
## Summary
- Adds `ModemPreset::Montreal` / `MONTREAL_PROFILE` (US915, 914.875 MHz, SF9, 125 kHz, CR 4/7) selectable at compile time via `PRNS_LORA_PROFILE=montreal`.
- Clamps TX power to each board's safe max (MT2 30 dBm, HV4/T096 28, T114 21, T-Echo/T1000-E/T-Beam 22) so the channel tag stays shared while local power fits the radio.
- Wires boot path and UI preset list; updates the MT2 build script to pass through the profile env var.

## Test plan
- [ ] `cargo test -p prns-core profile::tests`
- [ ] Build MT2 UF2 with `PRNS_LORA_PROFILE=montreal bash tools/build/hopspot-mesh-tower-v2.sh`
- [ ] Flash two boards on Montreal; confirm they hear each other on 914.875 / SF9
- [ ] Confirm UI can select Montreal and Reset LoRa Profile clears persisted non-Montreal settings where storage exists

Made with [Cursor](https://cursor.com)

PR #174, opened 2026-08-28
Mobile Hopspot: bootstrap RemoteControl identities for PrnsNodeRecipe

tonycowan:
## Summary
- Android and iOS Hopspot engines now load or generate persistent remote-control identity material and pass `RemoteControlService` into `PrnsNodeRecipe`, matching `prnsd` and embedded hosts.
- iOS persistence tests updated with a small test helper so they compile against the trunk API.

## Context
Trunk added a required `remote_control` field on `PrnsNodeRecipe`; mobile startup paths were the remaining gap and blocked Android/iOS builds.

## Test plan
- [x] `cargo test` in `personal-hopspot/mobile/ios/rust` (21 passed)
- [x] `cargo ndk -t arm64-v8a build --release` in `personal-hopspot/mobile/android/rust`
- [ ] `./gradlew installDebug` on a device and confirm Hopspot engine starts
- [ ] iOS engine smoke test on device/simulator

Made with [Cursor](https://cursor.com)

PR #172, opened 2026-08-28
Heltec V4-R8: PSRAM-backed run_core executor stack

tonycowan:
## Summary
- Add `heltec-v4-r8-boot` feature (enabled only on the Heltec V4-R8 board) to hand `run_core` off the main task onto a dedicated PSRAM-backed executor stack via vendored `esp-rtos::spawn_task_with_stack`.
- Non-R8 S3 boards keep the existing `Box::pin_in(ExternalMemory)` path unchanged.

## Test plan
- [ ] `cargo heltec-v4-r8` builds cleanly
- [ ] `cargo heltec-v4` (R2) still builds and uses the original boot path
- [ ] Flash V4-R8 and confirm boot completes past `run_core` without stack guard panic
- [ ] Confirm Wi-Fi associate and BLE/LAN interfaces come up on R8 hardware

Made with [Cursor](https://cursor.com)

PR #171, opened 2026-08-27
BLE Auto discovery groups (group_id) + status plumbing

tonycowan:
## Summary
- Partition BLE Auto discovery by configurable `group_id` (default `reticulum`): ADV v4 carries a 4-byte `sha256(group_id)` tag; SoftDevice / Trouble / macOS / Linux / Android advertise and filter on it; legacy v3 maps to the default group.
- Plumb `group_id` through config/`from_plan`, MeshTower `PRNS_BLE_DISCOVERY_GROUP`, Android persistence, and `prnsd status` (plus transport `software_version`).
- Harden isolation: macOS hard-rejects cross-group candidates; Android rejects inbound only when a scan proves a mismatch; classic Hopspot shows a read-only `grp` row (no cycle menu).
- Also includes nested Auto fleet peers in status (from #165) so island testing can see member peers under supervisors.

Still intentional follow-ups (not blocking radar): handshake-level group check; Dioxus UI to edit `group_id`; broader multi-device island lab.

## Test plan
- [ ] `prnsd` with `group_id = mt-leg-b` peers with Android Hopspot on the same persisted group; cross-group peers do not stay linked
- [ ] Default/`reticulum` Mac ↔ Android still peers without config
- [ ] `prnsd status` shows `Group : …` under Bluetooth Auto and `Software prnsd x.y.z`
- [ ] Classic Android BLE menu shows read-only `grp …` above peers (no Group action)
- [ ] MeshTower with `PRNS_BLE_DISCOVERY_GROUP` only sights matching island peers

Made with [Cursor](https://cursor.com)

PR #179, opened 2026-08-28
Fix WiFi Auto discovery and keepalive without IPv6 multicast

tonycowan:
## Summary
- **Discovery:** Carry DNS-SD over IPv4 mDNS (`MdnsMulticastFamily`) so peers can find each other when APs block IPv6 link-local multicast; fix Android/tokio scoped `fe80::` endpoints; treat embassy IPv6 discovery-group join as best-effort.
- **Keepalive:** Periodically unicast peering tokens to known peers on embassy (same cadence as tokio) so links survive after discovery when multicast beacons are blocked.
- Re-exports `MdnsMulticastFamily` from `personal-rns` alongside current trunk TCP rendezvous APIs.

Supersedes #173 (keepalive-only); that change is included here.

## Test plan
- [ ] Host/tokio: two peers discover each other on an IPv4-only or IPv6-restricted LAN
- [ ] Embassy/ESP32: AutoWifi discovery completes when IPv6 mDNS multicast is blocked
- [ ] Peers stay connected past the ~22s peering timeout without multicast beacons
- [ ] Android hopspot: WifiAutoLink still announces and resolves peers
- [ ] Smoke existing IPv6 link-local path still works where available


tonycowan:
Folded #173 into this PR (keepalive was already in the branch). #173 is closed as superseded.

tonycowan:
### Maintainer merge note

This PR is in **Wave 1** (independent of the other Wave 1 PRs) — safe to merge in any order with #140 / #174 / #179 / #181.

Then: #178 → #172 (stacked), then #171 last (stacked on #181).

Happy to rebase #171 after this lands if needed.

tonycowan:
Closing in favor of #229.

Same problem space (wifi-auto discovery/peering when IPv6 link-local mDNS multicast is unreliable), and #229 covers the newer host/Android/tokio hardening on current trunk, including scoped peer eligibility, host LAN inventory, macOS mDNS work, and embassy unicast peer refresh.

Note: #229 does **not** include this PR’s explicit ESP32 IPv4 mDNS multicast path (`MdnsMulticastFamily::Ipv4` / `wifi.rs`). If that approach is still needed for APs that block IPv6 LL multicast, it should be re-landed as a follow-up on top of #229 rather than merging this older branch.

PR #183, opened 2026-08-30
Stop hopspot-flash inheriting CARGO_TARGET_DIR

tonycowan:
## Summary
- `hopspot-flash` already strips `RUSTUP_TOOLCHAIN` and `RUSTFLAGS` from nested `cargo` so embedded builds do not inherit the parent invocation.
- Also strip `CARGO_TARGET_DIR`. Environments such as Cursor set it, which sends the ESP ELF to a sandbox target dir while the flasher still looks under `embedded/esp32/target/...`.

## Test plan
- [ ] From an environment with `CARGO_TARGET_DIR` set, run `hopspot-flash flash <board> --local-build` and confirm it finds the ELF under `embedded/esp32/target/...`
- [ ] Confirm a flash without `CARGO_TARGET_DIR` still works


KenAKAFrosty (approved):
Validated at current head 537617776: the helper now explicitly removes inherited CARGO_TARGET_DIR, matching the existing isolation of RUSTUP_TOOLCHAIN and RUSTFLAGS, and the regression test directly inspects the child command environment. cargo test --locked --manifest-path personal-hopspot/flasher/Cargo.toml passed (93 unit + 2 integration tests), and cargo fmt --manifest-path personal-hopspot/flasher/Cargo.toml -- --check passed. This fixes the mismatch between where nested Cargo writes embedded artifacts and where the flasher expects them.

PR #186, opened 2026-09-02
Keep the Hopspot LoRa card when Tune retags the radio

tonycowan:
## Summary
- Classify the Hopspot LoRa card by `InterfaceKind::LoRa`, not the boot-time `InterfaceId`.
- Tune and Save retag LoRa's id from frequency and modulation; matching the old id turned a live LoRa radio into a generic Peer card.

## Test plan
- [x] `cargo test --locked -p personal-hopspot-core screen::tests` on Darwin arm64 (98 passed, before this was split onto trunk)
- [ ] Flash Heltec V4, Tune/Save a Montreal (or other) channel, confirm the LoRa card stays LoRa instead of becoming Peer
- [ ] Confirm a real Peer interface still renders as a Peer card


KenAKAFrosty:
Thanks!  Ended up knocking this out back in 58cb73377  so just closing this now

PR #188, opened 2026-09-02
Ship 0.3.7-hotfix.5 ESP32-S3 entropy hotfix

KenAKAFrosty:
## Summary

- define 0.3.7-hotfix.5 directly on the complete current main source
- rebuild Heltec V4, Heltec V4 R8, and T-Beam Supreme with corrected ESP32-S3 TRNG ordering
- retain every unaffected hotfix.4 artifact byte-for-byte
- document full-chip erase as remediation for identities created by affected firmware
- close exhaustive pre-push license-policy gaps for standalone benchmark, tray, and fuzz graphs

## Validation

- complete repository pre-push CI-parity suite
- affected Xtensa release checks and two byte-identical local artifact builds
- core entropy and fixed-vector crypto tests
- Heltec V4 sparse flash and true reboot smoke
- Wi-Fi association and DHCP on both boots
- exactly one configured outbound TCP client connected after reboot
- LoRa GC1109, BLE, ESP-NOW, persistence, watchdog, and runtime entropy healthy
- no packet refusals, queue failures, panics, or resets observed

No selective backport is used. This branch starts at protected main f28ba636462f2255227e51cc19c7a9d6dd41343e and adds release metadata, documentation, and policy corrections only.

### DIFF
--- a/about.toml
+++ b/about.toml
@@ -26,6 +26,9 @@ accepted = ["LicenseRef-Nordic-SoftDevice"]
 [serialport]
 accepted = ["MPL-2.0"]
 
+[option-ext]
+accepted = ["MPL-2.0"]
+
 [nrf-softdevice-s140.clarify]
 license = "LicenseRef-Nordic-SoftDevice"
 override-git-commit = "47d6121c6e823120e8b883a7ac75f44ce7daa3aa"
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -3,6 +3,7 @@ name = "personal-reticulum-suite-benchmarks"
 version = "0.0.0"
 publish = false
 edition = "2021"
+license = "MIT OR Apache-2.0"
 autobins = false
 
 # Standalone, like `validation/fuzz/`: an empty `[workspace]` detaches this from the root so
--- a/deny.toml
+++ b/deny.toml
@@ -44,6 +44,9 @@ exceptions = [
     # tray-icon 0.24.1 reaches platform configuration through dirs 6.0.0 and this small
     # MPL-licensed option adapter. Keep the exception pinned instead of allowing MPL globally.
     { allow = ["MPL-2.0"], crate = "option-ext@0.2.0" },
+    # The non-shipping fuzz harness links LLVM's libFuzzer runtime. NCSA is permissive,
+    # but keep its acceptance pinned to the exact test-only package instead of widening products.
+    { allow = ["NCSA"], crate = "libfuzzer-sys@0.4.12" },
 ]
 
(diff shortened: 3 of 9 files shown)


jackspace:
Hey Ken, I ran the R8 follow-up from the hotfix.5 deferred hardware list on one of my Heltec V4-R8 boards, using the signed 0.3.7-hotfix.5 release and the hopspot-flash CLI.

Short version: 8 of the 10 checks and scenarios in the hotfix spec pass, one I can only show as a code path, and LoRa readiness is still partial.

This was the signed release as tagged, so it doesn't include #172.

Esptool reports 8 MB of embedded PSRAM on this board, which is how I told it apart from an S3R2.

Doctor can't separate the 16 MiB boards, and it says so itself.

The manifest signature verified against the release key, and the flasher checked all three R8 parts against the signed manifest.

I read the whole flash twice as a backup, did a full chip erase, then flashed with `hopspot-flash flash heltec-v4-r8 --version 0.3.7-hotfix.5 --allow-downgrade --port COM16 --yes --monitor`.

- Full-erase first boot, post-flash boot and second boot all reached `core1.executor.ready` with no panic and no unexpected resets.

- The first boot logged `node identity created` and `Bluetooth identity created`, and the identity region no longer matched the pre-erase backup.

- After two more boots the identity region was byte-identical, and neither boot logged a new identity.

- For stable runtime I watched for 16 minutes with the port held open, saw no reset after the watch's own opening one, and the four `state persisted` lines came about 300 seconds apart.

- There's no log line for the TRNG coming up before identity generation, so all I can point at is the code path in `s3/entropy.rs`, where the TRNG source is created before `bootstrap_s3_identities` runs.

- The KCT8103L FEM was detected on every boot, but I haven't caught an on-air frame I can pin to this board yet, so I'm not counting LoRa readiness.

One question: the CLI never does a full-chip erase, so I used `espflash erase-flash` for that step.

Is espflash the erase you expect for this scenario, or would you rather the flasher had one?

Happy to share the logs if they'd help.

jackspace:
A quick follow-up on the R8 note above: I've now seen LoRa readiness on air, on the same board and the same signed 0.3.7-hotfix.5 build.

I put an RNode on the default 915 MHz profile (SF9, BW 250 kHz, CR 4/5) and listened for the board's own node destination, then switched the board between powered and unplugged in five-minute windows.

In all three powered windows, each Announce press was heard directly at hops 1 about a second later.

Both unplugged windows were completely silent, with the RNode's receive counter flat.

I wouldn't call the silent windows a controlled negative, though, because the RNode heard nothing at all while the board was unplugged, so nothing shows the receiver stayed live.

So that row moves to observed, which puts it at 9 of the 10 in the hotfix spec, with the TRNG check still a code path only.

Happy to share the logs if they'd help.

PR #197, opened 2026-09-08
Add a self-contained Prns app

evelant:
This draft adds a self-contained Prns app under `applications/` for iOS and Android. It supports identity creation/import, contacts, RemoteControl pairing and authenticated checks, node-address sharing, and small direct LXMF messages with a persistent mailbox.

The goal is to develop the app alongside Prns while keeping product policy, storage and UI outside the core. Feedback on that boundary and sensible landing steps is welcome.

## Architecture

The Expo UI uses generated UniFFI TypeScript, Swift and Kotlin bindings over one process-owned Rust supervisor and shared library image. Rust owns protocol workflows and durable state; platform code owns permission, storage and lifecycle admission. Native startup and iOS restoration can precede JavaScript. The `no_std` core has no UniFFI dependency.

The React Native generator/runtime is pinned with recorded patches and reproducible artifacts. See the [binding guide](https://github.com/evelant/Prns/blob/prns-app/applications/prns/native-composition/bindings/README.md).

The existing UniFFI 0.31.2 dependencies have exact-version MPL-2.0 exceptions and matching notices for the app's iOS and Android builds. The repository-wide license allowlist is unchanged.

## Current upstream integration

Updated against trunk `c4fd54dfd`, including #232. The app now projects all remote-control request kinds and displays Operator or Administrator authority before pairing approval. Existing board pairing grants are not silently expanded. The new inventory, configuration, Wi-Fi and access-management screens are **planned, not implemented**; the [implementation plan](https://github.com/evelant/Prns/blob/prns-app/applications/docs/remote-control-expansion.md) covers their capability-driven UI and failure/recovery behavior.

Reusable core/platform contributions remain separate PRs; this app branch temporarily includes their integration. #198, #213, #216, #221 and #224 are merged, and #214 is superseded. The remaining follow-up fixes are published separately.

## Validation and limits

- Fresh detached app checks at `85cbd3754` pass: 136 native unit tests, four integration tests, 49 SDK tests, 223 UI tests, generated-output checks, Expo Doctor 21/21, web export and two-way Python LXMF exchange. Separate fresh Swift lifecycle, startup-queue and release-symbol checks also pass.
- The normal publication hook passed, including all 24 host workspaces, 14 configured firmware profiles and the selected Miri, target-ISA and integration checks. The pinned core revision was also fetched and verified from the declared public repository.
[...]

PR #198, opened 2026-09-08
Gate core tests and examples on the features they require

evelant:
Some core tests and the Turbo example are selected in builds that do not enable the features they need, causing compilation failures.

Core changes needed by the Prns app must preserve support for builds without `std`. These failures blocked the alloc-only regression checks; correct feature guards make those checks usable again.

Require `std` for the Turbo contention and acquisition-simulation tests and the `subghz_turbo_evidence` example. Require the appropriate allocation/offload features for the affected resource tests. Production code and test assertions are unchanged.

## Tests

Passed:

- Alloc-only validation, including strict Clippy, unit tests, crypto tests, and doctests
- All four affected resource tests with offload enabled
- Turbo tests and evidence example with `std`
- All-target compilation checks for the affected feature combinations

## App integration

The earlier firmware-size issue is fixed in the app integration (#197). App commit `f3755a977` passes all 14 configured firmware profiles and the normal local publishing gate; T-Echo S140 v7 has 624 bytes of flash headroom, with limits and layouts unchanged. See the [September 10 validation](https://github.com/evelant/Prns/blob/f3755a9773cc7a305a8811ba679e0638f96ab52d/applications/checkpoints/2026-09-10-app-publication.md).

These results apply to the integrated app, not a fresh build or CI run of this PR. The branch-specific checks above remain separate.

### DIFF
--- a/prns-core/Cargo.toml
+++ b/prns-core/Cargo.toml
@@ -25,6 +25,10 @@ include = [
 [lib]
 path = "src/lib.rs"
 
+[[example]]
+name = "subghz_turbo_evidence"
+required-features = ["std"]
+
 [features]
 default = ["std", "std-sync-host", "resource-work-offload"]
 # `zeroize/alloc` rides our own `alloc` gate so `Zeroize for Vec<u8>` is present by declaration rather than incidental dependency feature unification. The host identity vault's keyring backend zeroizes the secret bytes the OS store hands back.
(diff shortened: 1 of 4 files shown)


KenAKAFrosty:
Thanks a ton for all the care and fixes you've submitted! I'm starting to work through them slowly now

PR #199, opened 2026-09-08
Await USB Auto test input without busy polling

evelant:
Upstream now supplies the shared wake signal required by the USB Auto fixtures.
Keep the remaining test improvement: await the inbound grant lane with the
existing two-second deadline instead of repeatedly yielding and polling it.
Handshake, payload and lifecycle assertions are unchanged; production code is
untouched.

The Prns app's transport and restoration work relies on these host-side tests.
This keeps their input wait event-driven without introducing an app dependency.

Rebased on current trunk. All 17 focused USB Auto tests and the normal local
pre-push checks pass, including the 14-profile firmware matrix. Current GitHub CI
is reported separately. App integration: #197.


KenAKAFrosty:
Sorry for the delay but this kind of thing is VERY welcome. I plan to get caught up on many of these PRs today. Thank you very much!

PR #201, opened 2026-09-08
Share Host snapshot construction and JSON serialization

evelant:
Extract Host snapshot construction into a shared crate so other Hosts can reuse
it without copying native Host internals. Preserve fleet grouping, route
ownership, caller-supplied capabilities and persistence information. JSON output
keeps contract-required `u64` values as decimal strings and rejects unsafe
JavaScript integers.

The Prns app needs this shared projection and serializer to keep its Rust bridge
aligned with the public Host contract. The crate remains independent of the app.

Rebased on current trunk, refreshed generated notices and updated fixtures for
the expanded interface inventory. The shared snapshot test, three JSON tests,
24 native Host tests and strict native Clippy pass. The normal local pre-push
checks also pass, including the 14-profile firmware matrix. Current GitHub CI is
reported separately; no new hardware result is claimed. App integration: #197.


KenAKAFrosty:
(deferring the JSON serializer for now but can def come back to that later)

PR #208, opened 2026-09-08
Add central-only CoreBluetooth Auto support

evelant:
Allow Bluetooth Auto to run with only a central manager, without publishing peripheral services. Keep the existing dual-role behavior for other callers.

The Prns iOS app uses accessory-selected outgoing connections and needs this role without enabling unrelated peripheral services. Central-only does not mean foreground-only.

Depends on #207 and #202; [own changes](https://github.com/evelant/Prns/compare/16713c9a9b494385ac878644d58f5d68b14f8a76...5750e224199e24f9f7536b65f287e284a91aa8d4). App integration: #197.

Rebased on trunk `c4fd54dfd`. Validation on `5750e2241`: 81 FFI tests pass with and without logging (one hardware test ignored per run), 14 Bluetooth-interface tests pass, and strict FFI Clippy and iOS Rust compilation pass. The normal publishing gate passed, including the configured firmware resource matrix, Miri/target-ISA checks and selected SDK lanes. Fresh physical-device qualification remains separate.

### DIFF
--- a/audits/unsafe-snapshot.json
+++ b/audits/unsafe-snapshot.json
@@ -16840,11 +16840,11 @@
       "name": "prns-ffi",
       "source": "prns-ffi",
       "unsafe": {
-        "blocks": 124,
+        "blocks": 138,
         "externs": 3,
         "functions": 0,
         "impls": 22,
-        "tokens": 185
+        "tokens": 199
       },
       "version": "0.3.7"
     },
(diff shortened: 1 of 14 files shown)


evelant:
Closing because the Prns app now uses ordinary dual-role CoreBluetooth for phone-to-phone networking, so the accessory-selected central-only path is no longer needed.

State restoration remains supported by merged #202/#207 and the #218 recovery follow-up. The useful scan diagnostics remain in #209, which no longer depends on this PR. The branch is retained for reference.

PR #213, opened 2026-09-10
Preserve split-resource identity after a segment completes

evelant:
A split resource can lose its chain identity when one segment completes before
the next segment's offloaded build is reserved. That continuation takes the live
lane instead of the staged lane, but advertises its own hash as the original
resource hash. The receiver cannot match it to the existing assembly, so the
sender can exhaust its advertisement retries. For request responses, that
failure also closes the link.

Copy the existing assembly's original hash when a live continuation build lands,
matching the inline and staged paths. No retry, timeout, queue or wire-format
changes are needed.

This surfaced while preparing the app's upstream branches (#197): the
large-response TCP check also fails on current trunk, independently of USB or
Bluetooth. The app needs reliable large responses and a sound foundation for
future Resource support; this does not add large-message support to the app.

The regression proves segment one before reserving segment two, checks the
continuation's decrypted advertisement, and settles both segments. It fails
before the correction and passes afterward. On Linux arm64, 1,887 core tests,
25 integration tests and 50 consecutive runs of the original split-response
test pass without diagnostics. Three existing profiling tests and one manual
interop server remain ignored. Strict host core Clippy, an embedded-target core
check and the canonical unsafe inventory also pass.

All fourteen configured firmware profiles pass; T-Echo S140 v7 retains 472 bytes
of flash headroom. No memory layout or capacity changes are included.

The initial macOS Bluetooth CI job hits the existing USB test-fixture compilation
error addressed by #199; this change does not touch that code.

### DIFF
--- a/prns-core/src/routing/links/resources/send.rs
+++ b/prns-core/src/routing/links/resources/send.rs
@@ -670,6 +670,12 @@ impl<S: StorageLayout> EngineState<S> {
         let state = *self.outgoing_resources.state(index);
         if state.segment_index == 1 && state.total_segments > 1 {
             self.outgoing_assemblies.begin(reservation.link_id, hash);
+        } else if state.segment_index > 1 {
+            // The predecessor may have proved before this build was reserved, so a live
+            // continuation needs the same chain identity as a promoted staged build.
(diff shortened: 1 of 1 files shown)


KenAKAFrosty:
FANTASTIC edge case find & fix, thank you!

PR #214, opened 2026-09-10
Refresh notices for embedded tooling dependencies

evelant:
The locked dependency graphs gained `personal-hopspot-builder`,
`personal-hopspot-memory`, `personal-hopspot-resources`, and `rustc-demangle`
after the last notices refresh. Regenerate the bundle to record their use of
existing MIT license texts.

This removes an inherited notices-check failure encountered while preparing the
Prns app (#197) and its USB and iOS Bluetooth prerequisites, while keeping the
generated-file update separate from those changes. No dependencies, license
policy or generator behavior change.

Validation: reproduced the current-trunk failure and its exact match with the
earlier CI diff; regenerated with cargo-about 0.9.1 and Rust 1.96.0; an independent
fresh-cache check reproduced the file byte-for-byte. All five generator tests
and license-policy parity pass. The canonical checks ran on macOS against the
28 registered release graphs.


evelant:
Closing as superseded: current trunk includes all four notice additions from this PR, including the Hopspot tooling crates and rustc-demangle. The app no longer needs a separate notice prerequisite. Thanks for incorporating these upstream.

PR #221, opened 2026-09-13
Refresh all affected wake schedules when a link closes

evelant:
Closing a Link can settle requests, close channels and abort pairing. Return fresh receipt, channel and pairing wake schedules with the command's delta, so the runtime does not retain deadlines for removed state.

The Prns app closes Links when a remote check is cancelled or the node stops. Correct wake bookkeeping lets that cleanup finish without waiting on obsolete work. Regressions start with real pending work, compare the delta with the engine's authoritative schedules, and preserve deadlines for work on other Links.

This branch already includes trunk `79050535f`; no history rewrite was needed. Earlier focused, no-std and firmware validation remains historical. No fresh firmware margin or physical-device result is claimed for this refresh.


KenAKAFrosty:
(loving these extra care and diligence tune-ups. genuinely valuable ty!)

PR #225, opened 2026-09-13
Wait for MTU completion before discovering services

evelant:
Wait for the MTU callback or a synchronous request rejection before starting Android service discovery. Releasing a local gate after 750 ms does not cancel Android's request and can overlap the two operations. Keep the eight-second startup timeout, reject duplicate completions and stop further operations after close.

This surfaced during Prns app (#197) development on Android 10. Personal Hopspot has the same sequence; this fixes its separate adapter too. It is not a general fix for vendor GATT status-133 errors.

Rebased on current trunk. All eight operation-state unit tests pass. The normal local pre-push checks also pass, including all 14 firmware profiles within their unchanged limits. Current GitHub CI is reported separately. The earlier Android service smoke is historical, not a fresh result for this head; no new physical-device qualification is claimed.

### DIFF
--- a/personal-hopspot/mobile/android/app/src/main/java/org/personal/hopspot/BleLink.kt
+++ b/personal-hopspot/mobile/android/app/src/main/java/org/personal/hopspot/BleLink.kt
@@ -27,13 +27,12 @@ import android.bluetooth.le.ScanSettings
 import android.content.Context
 import android.os.Build
 import android.os.ParcelUuid
+import android.os.SystemClock
 import android.util.Log
 import java.nio.ByteBuffer
 import java.util.UUID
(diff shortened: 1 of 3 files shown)


KenAKAFrosty:
Oh yeah very very very welcome fix-ups here.  Have some medium-future plans to come back and properly state-machine-ify a lot of work in the interfaces and the core; at that point a *lot* less of these paper cuts should show up in the first place.

But that's exactly why it's VERY helpful and encouraging to have your help already this early on, smoothing over these paper cuts. Thanks!

evelant:
@KenAKAFrosty FYI I'm doing a bit of refinement on some of the dependent / draft PRs right now. Thanks for merging this stuff!

PR #227, opened 2026-09-14
Seeed SenseCAP Solar Node P1/P1-Pro target

m1ndgames:
**Preamble:**
I was looking for a solution to run reticulum on my Seedstudio SenseCap Solar Node, and all i could find was this projects Roadmap. As i barely speak Rust and im already overloaded with personal projects, i pointed claude (Opus 5 xHigh) at the issue.

I am aware that contributors should own their code, but i need to be transparent on this: I am not able to further support this, on the other hand i did not want to keep a working feature to myself.

**I am fine with this PR being rejected**, but maybe someone else with more time on their hands wants to pick it up where i left it. (This is also the reason i did not remove the over-explanatory claude code comments)

My testing setup:
Seeed SenseCAP Solar Node P1/P1-Pro as target
Raspberry with attached Heltec v4 rnode

**PR message**
Adds the Seeed SenseCAP Solar Node P1/P1-Pro as an nRF52840 Hopspot board target and promotes it from Tier::Roadmap to Tier::BringUp.                                                                                                                                                                
                                                                                                                                                                                                                                                                                                      
The board is a carrier for a XIAO nRF52840 Plus and a Wio-SX1262, so it is structurally a T1000-E with an SX1262 in place of the LR1110 — same bootloader, same resident S140 7.3.0, same 0x27000 application base — and it reuses that proven flash geometry. Pin map cross-checked against the Meshtastic and MeshCore variants and Seeed's block diagram, which agree on every pin.

Also adds DEFAULT_868_PROFILE, since every headless board was previously pinned to DEFAULT_915_PROFILE. It transmits at 14 dBm because Region::Eu868 caps there.
                                                                                                                                                                                                                                       
Hardware receipt in validation/qualifications/. Transmit was decoded by stock Python RNS 1.5.4 as a valid Reticulum packet; receive was confirmed with the peer's LAN interface disabled and the host holding no cached paths, so the two-hop route learned through the board's own transport instance could only have arrived over LoRa.                                                                                                                                                                                                                                                          

Deliberately not included: release/flash/boards.json and flash-manifest entries, since those assert a signed release and hardware acceptance a contributor can't supply. Follows the MeshTower V2 precedent — firmware and build task, no catalog entry.


tonycowan:
thanks @m1ndgames !

KenAKAFrosty:
Hey @m1ndgames , thanks so much for sending this over our way! My capacity's a bit limited right now so forgive the delayed response, and it'll take a few a days at least before I get a chance to review this, but I really appreciate you taking the time to share this

jackspace:
I put this branch on a P1-Pro here, with two differences from your setup that might be useful data points. My unit shipped with the stock XIAO nRF52840 bootloader (0.9.2, volume `XIAO-BOOT`, S140 7.3.0), not the SenseCAP OTA-fix one, and the 0xADA52840 family ID was accepted fine. And I'm in the 915 band, so I switched the headless profile to `DEFAULT_915_PROFILE` and left everything else as you had it.

Judged by stock Python RNS 1.4.2 on an RNode (Heltec V3, 915.0 MHz / 250 kHz / SF9 / CR5), with only the Solar Node and the RNode on the bus: the board's own boot announce decoded at 1 hop; a prnsd announce over USB Auto decoded at 2 hops via the board's transport identity; and `rnprobe` from the stock side round-tripped through the board to prnsd and back, 991 ms over 2 hops, 4 of 4, 0% loss. I also built a variant that holds the L76K disabled at boot, since nothing headless consumes a fix and it's a solar budget; boot announce at 1 hop and a 1.27 s round trip. I then repeated the relay and probe checks on that variant against RNS 1.5.4 with fresh storage, so every path had to be learned over the air: the prnsd announce arrived at 2 hops via the board, and two probes came back in 1.14 s and 945 ms. Happy to share the L76K diff if it's wanted, and it's a judgment call, so no strong opinion.

One thing you'll hit in CI: the branch fails `validation/hygiene/fmt-docs.sh` on five of its own files (`personal-hopspot/core/src/lib.rs`, the nrf52840 `boards/mod.rs`, both `sensecap_solar_node` board files and `runtime/bootloader_entry.rs`). main passes under the same rustfmt, so it's the PR. It's a pure `cargo fmt` of those two packages and it applies cleanly on your head, so you can run it yourself or I'm happy to share the patch. Firmware clippy with `-D warnings`, the host tests on the crates you touched, and host clippy are all clean.

PR #237, opened 2026-09-18
Publish Wi-Fi peer link-locals in the existing radio slot

tonycowan:
## Summary
- Pack an EUI-64 Wi-Fi station MAC into the existing radio indication (tag `0x08` plus 6 bytes). `RadioIndication` stays 10 bytes, and the last byte of the 8-byte wire slot stays zero.
- A controller can rebuild the `fe80::/64` address from that MAC. Privacy IIDs do not fit this slot and stay unpublished.
- T096 still links. On this Mac the resource report shows `static_section_bytes` 143208 and the runtime-stack check passed.

## Test plan
- [x] `cargo test -p prns-core --locked --lib` for the radio-slot round trip, privacy-IID rejection, layout, and vitals publish
- [x] `env -u HOPSPOT_WIFI_PASSWORD -u HOPSPOT_WIFI_SSID ./tools/prns build embedded resources report --target t096`
- [x] Pre-push CI parity on darwin aarch64, including embedded ISA for thumbv7em, riscv32imac, and xtensa-esp32s3
- [ ] Confirm a peer inventory rebuilds the link-local only when the station address is EUI-64


tonycowan:
Closing this. The radio indication stays a radio fact, and the interface card group string stays a group name. A peer name decoration has to use an identifier both sides already have, not a reused field.

PR #238, opened 2026-09-19
benchmarks: record every native module the compiled reference loads

jackspace:
I went looking at the benchmark reference loader to answer a question about our published numbers,
and found that the proof record can't answer it.

`compiled_reference.py` builds the full set of native RNS extension modules that loaded, checks the
set isn't empty, and then records `native_modules[0]`. So the gate confirms that *a* native module
is present, never which one or how many, and the `suite.json` it writes keeps only the first. After
a run there's no way to tell a reference where the protocol machinery compiled from one where only
the package `__init__` did.

On this bench it's the second one. At Cython 3.2.8 the compiled reference produces **1 native
extension module and 43 interpreted ones**. The native one is `RNS` itself. `RNS.Transport`,
`RNS.Link`, `RNS.Packet`, `RNS.Identity`, `RNS.Destination`, `RNS.Resource` and `RNS.Reticulum` are
all interpreted, along with the whole of `RNS.Cryptography` and every `RNS.Interfaces` module.

That's the same answer on four setups:

| host | arch | python | compiler | native / interpreted |
|---|---|---|---|---|
| Windows 11 Pro 26200 | x86_64 | 3.13.6 | MSVC | 1 / 43 |
| Ubuntu 24.04 | x86_64 | 3.12.3 | gcc 13.3.0 | 1 / 43 |
| Kali rolling (Pi 4B) | aarch64 | 3.14.6 | gcc 15.2.0 | 1 / 43 |
| Ubuntu 24.04, RNS 1.5.4 | x86_64 | 3.12.3 | gcc 13.3.0 | 1 / 45 |

It isn't new either. The `suite.json` files already published on trunk record a single top-level
module for macOS, Linux and Windows alike, so this has been visible in our own results the whole
time, just not in a form anyone could read.

The fix is small. The data was already being computed and then thrown away one line later, so this
records all of it: `native_modules`, `native_module_count`, and the interpreted complement as
`interpreted_modules` and `interpreted_module_count`. I kept `native_module` so the suites published
so far stay comparable.

The second commit is the descriptor wording. Both `rns-*-compiled` entries said runs "are rejected
unless RNS.compiled is true and native extension modules are loaded". Both conditions really do
hold, but the sentence reads as an assurance that the reference is a native build, and that's more
than the check can promise. They now say what the gate verifies, what the inventory measured, and
when.

I've deliberately left the gate alone. It still accepts the run. Whether it should start asserting
that the protocol modules are native, or whether the label should change instead, is a call about
what the comparison is meant to mean, and that's yours rather than mine.

To see it for yourself:

```
cd benchmarks/reference
python -m venv .venv && .venv/bin/python -m pip install -r requirements.lock
.venv/bin/python -u compiled_reference.py --verify-only
```

The `REFERENCE_PROOF` line now carries the counts and both lists.

Gates green on this branch at `6e28f8798`: `VALIDATION_REGISTRY_OK`, `TOOLING_REGISTRY_OK`,
`FMT_DOC_CHECK_GATE_OK`.


tonycowan:
Thanks Jack!

KenAKAFrosty:
Oh nice, thanks Jack!  When this was first set up, the intent was to keep it comparable to how it's ran in the reference implementation.  Looks like just a few weeks ago they also updated to similar, so this is a very welcome change!

PR #239, opened 2026-09-19
feat(hopspot): add XIAO ESP32-S3 Wio-SX1262 support

alearcy:
## Summary

Adds support for the Seeed XIAO ESP32-S3 with the Wio-SX1262 LoRa expansion as a headless Personal Hopspot target.

This PR includes:

- XIAO ESP32-S3 / Wio-SX1262 board definition, pin mapping, SX1262 frontend control, and 8 MiB flash layout.
- Headless display runtime, with no display-specific UI assumptions.
- Flash manifest and flasher catalog support for `xiao-esp32s3-wio-sx1262`.
- Persistent LoRa profile provisioning through:

  ```console
  cargo prns-esp32-flasher configure xiao-esp32s3-wio-sx1262 ...

- Optional persistent node auto-announce interval and public node display name.
- Documentation for flashing and provisioning the board.

configure writes and verifies only the two radio-profile flash pages. It preserves firmware, Wi-Fi provisioning, identities, and retained node state.

## Validation

- cargo test --locked -p personal-hopspot-core — 168 tests passed
- cargo test --locked -p hopspot-flash — 100 tests passed
- cargo fmt --all -- --check
- XIAO ESP32-S3 release cross-check passed:

  cargo +esp check --release --locked \
    -p hopspot-xiao-esp32s3-wio-sx1262 \
    --target xtensa-esp32s3-none-elf \
    -Zbuild-std=core,alloc

- Hardware flash validation was performed on the XIAO ESP32-S3 / Wio-SX1262 target.

## Notes

The registered full ESP firmware gate is Linux-only, so it was not run on the macOS development host. The XIAO-specific release cross-check above was run successfully.


alearcy:
I added the LED support

PR #240, opened 2026-09-20
interfaces: match the MTU tier boundaries RNS uses today

jackspace:
`hardware_mtu_for_bitrate` mirrors RNS 1.4.2, and RNS changed the comparison in 1.5.1. The tier values are still identical, but every boundary went from `>` to `>=`, so a bitrate sitting exactly on a boundary now resolves one tier higher in RNS and one tier lower in Prns.

That would be a footnote except for which boundary it is. 10,000,000 is `TCPInterface.BITRATE_GUESS`, what both `TCPClientInterface` and `TCPServerInterface` assume when a config names no bitrate, and it has not changed in any release I looked at. So the divergence lands on the most common TCP configuration there is:

| | at 10 Mbps | HW_MTU |
|---|---|---|
| RNS 1.4.2 | `> 10_000_000` | 8192 |
| RNS 1.5.0 | `> 10_000_000` | 8192 |
| **RNS 1.5.1** | `>= 10_000_000` | **16384** |
| RNS 1.5.4 | `>= 10_000_000` | 16384 |
| Prns, before this | `10_000_001..` | 8192 |

I found it by accident. I added an RNS 1.5.4 arm to the benchmark matrix alongside the pinned 1.4.2 one, and the relay cells aborted on the harness's own assertion that the reference's reported policy matches `hardware_mtu_for_bitrate`:

```
assertion `left == right` failed: RNS 1.5.4 (interpreted) reported a TCP policy
that does not match raw-transport-throughput
  left:  RelayTcpPolicy { bitrate_bps: 10000000, mtu_bytes: 16384 }
  right: RelayTcpPolicy { bitrate_bps: 10000000, mtu_bytes: 8192 }
```

Same bitrate on both sides, different MTU. Every 1.5.4 participant in that run reported 16384 and every 1.4.2 participant reported 8192, across 42 process starts, so it is not a one-off. I then read `optimise_mtu` out of both pinned venvs rather than trusting the harness, and bisected it against RNS's release tags to 1.5.1.

This change makes the tiers inclusive to match, and adds the test that was missing. Nothing asserted the boundaries before, which is why this could drift quietly, so the test checks every tier from three sides: one below, exactly on it, and one above. The exactly-on-it column is the part that moved.

**What it affects.** `hardware_mtu_for_bitrate` has two callers: `MtuPolicy::OptimizedFromBitrate::resolve` and the benchmark's expected relay policy. Five interface families resolve MTU from bitrate and are affected at boundary bitrates, `tcp`, `udp`, `websocket`, `wifi_auto` and `shared_instance`. Everything else uses `MtuPolicy::fixed` and does not change. Away from the boundaries nothing moves at all.

This is a behaviour change at those boundaries rather than a pure bug fix, so it is your call whether matching current RNS is what you want here. I have written it the way that makes Prns agree with the RNS people are actually running today. You mentioned on #238 that the intent was to keep the reference comparable to how RNS actually runs it, and this felt like the same thing one layer down.

**Why nothing caught it, which I think is the more useful part.** Every RNS in the repo predates the change. `benchmarks/reference/requirements.lock` pins `rns==1.4.2`, and `validation/interop/requirements.txt` pins `rns==1.5.0`, which is the last release before it. The pin is one release short, so no suite we have could have seen it. The doc comment on the function said `RNS 1.4.2 Interface.optimise_mtu` and was accurate about what it mirrored the whole time.

Verified on x86_64-unknown-linux-gnu, Python 3.13.15, against RNS 1.4.2 and 1.5.4 installed from the pinned locks. I have not tested macOS.


tonycowan:
Legend.

jackspace:
Looks like this landed in d253e964f with the RNS 1.5.4 benchmark alignment: same inclusive boundaries, and a test that pins each exact threshold. That's the same fix with its own regression test, so I'll close this out. Thanks!

